PR #95, opened 2024-06-14
Add write permissions for doc workflow.

xingjian-zhang:
## Description

As titled.

### 1. Motivation and Context

- `doc` action failed in https://github.com/TRAIS-Lab/dattri/actions/runs/9521276731/job/26248328817#step:6:117.
- Reference: https://stackoverflow.com/a/76645503.

### 2. Summary of the change

As titled.

- [x] N/A: No test will be added (please justify)
- [ ] ...

It is hard to test locally as this is a Github action.


jiaqima (approved):
LGTM

PR #96, opened 2024-06-15
[Doc] add api docs

TheaperDeng:
### 1. Motivation and Context
This PR is to enable the api doc in our document. Some of the display issue can be fixed by the developer of each function.

- [x] Document test: If you added an external API, then you should check if the document is correctly generated.

image


xingjian-zhang (approved):
LGTM.

Issue #90, opened 2024-06-01
Model checkpoint alignment between `dattri_retain` and previous re-training


tingwl0122:
For Cifar-2 (and Cifar-10), the current checkpoints generated by our retrain script have some problems.
While our dattri algorithm can successfully be run on our previous re-training checkpoints

Issue #89, opened 2024-06-01
`dattri_retrain.py` need to set `start_id` from `partition` argument


tingwl0122:
Add one more line at the bottom:
```python
    if args.mode == "lds":
        kwargs = {}
        kwargs["num_subsets"] = int(args.partition[1]) - int(args.partition[0])
        kwargs["total_num_subsets"] = int(args.partition[2])
        kwargs["start_id"] = int(args.partition[0])
```

PR #97, opened 2024-06-25
[dattri.benchmark] Correcting the performance of (cifar2) benchmark experiment

tingwl0122:
## Description
This PR fixed the "lower-than-usual" LDS performance issue for cifar2-resnet9 benchmarking experiment.

### 1. Motivation and Context
We found the LDS of cifar2-resnet9 benchmarking is problematic, and we originally thought that it is the data's problem. (the retrain script is actually correct.) 

However, there are various minor issues within the entire benchmarking pipeline (the pipeline is detailed in `scripts/cifar2_resnet9_benchmark.py`) as well as a major design problem within TRAK and TracIn that makes the performance worse. 

To be short, if we first extract multiple model checkpoints into a list and feed it into `TRAKAttributor` or `TracInAttributor` as one of the inputs, then the gradient computation regarding these checkpoints will be wrong. The solution is just to read each checkpoint in the `Attributor` in a for-loop style and compute gradients one-by-one. Hypothesis is that pre compute/store checkpoints will destroy the computational graph and thus the gradient computation while we haven't figured out the true reason.

This PR will solve #90. Also, another one-line change in `scripts/dattri_retrain.py` will solve #89.

### 2. Summary of the change
- `model_utils/retrain.py`: add seed control
- `metrics/ground_truth.py`: modify output shape of `target_values` by replacing `dataloader.dataset` with `dataloader.sampler`
- `benchmark/datasets/cifar2/cifar2_resnet9.py`: replace the original **average loss** by  **per-example loss** to generate correct model output for `calculate_lds_ground_truth` function
- `algorithm/TRAK, Tracin`: modify the model checkpoint reading and add `model` as another function input to read checkpoints within the attributors
- `scripts/cifar2_resnet9_benchmark.py`: add this script to run benchmark experiment for cifar2
- `scripts/dattri_retrain.py`: set start_id from partition argument to solve #89 

TODOs (not necessarily in this PR):
- [x] correct the IF and RPS part in `scripts/cifar2_resnet9_benchmark.py` (useful for smaller experiments)
- [ ] modify other benchmark experiments' loss definition to per-example loss.
- [ ] create a unified benchmark script (now it is customized for cifar2-resnet9)
- [x] check other experiments' performance using this new version of TRAK and TracIn


dattri/algorithm/tracin.py
```diff
@@ -48,9 +47,9 @@ def f(params, image, label):
                     return loss(yhat, label)
                 ```.
                 This examples calculates the loss of the model on input data-label pair.
-            params_list (Union[List[dict], List[str]]): The parameters of the target
-                function. The input should be a list of dictionaries, where the keys
-                indicate the name of a parameter and the value is the parameter tensor.
+            model (torch.nn.Module): The PyTorch model to be attibuted.
+            params_list (List[str]): The parameters of the model, should be a list of
```
jiaqima, line 51: The name `params_list` does not align well with its content now

jiaqima:
> To be short, if we first extract multiple model checkpoints into a list and feed it into TRAKAttributor or TracInAttributor as one of the inputs, then the gradient computation regarding these checkpoints will be wrong. 

Is it possible to have a minimal toy example that reproduces this problem? E.g., on a linear module with multiple checkpoints. And show that grad_func will go wrong.

dattri/algorithm/tracin.py
tingwl0122: will make it `checkpoint_list`.

tingwl0122:
> > To be short, if we first extract multiple model checkpoints into a list and feed it into TRAKAttributor or TracInAttributor as one of the inputs, then the gradient computation regarding these checkpoints will be wrong.
> 
> Is it possible to have a minimal toy example that reproduces this problem? E.g., on a linear module with multiple checkpoints. And show that grad_func will go wrong.

Sure, but I guess for linear module, the problem may not be that serious for some reasons. As our benchmark experiment numbers recorded in the paper, MNIST+LR/MLP do not have this "LDS performance drop" even though the implementation (especially TRAK w/ multiple checkpoints) is the same across different settings.

tingwl0122:
Hi @jiaqima  @TheaperDeng 
I confirmed that this "gradient issue" does not happen on MNIST exps (for both lr and mlp model), but will appear on CIFAR2 exp. These are the smallest models that we can have exact gradient computation and store those gradient results for examination.

The related test cases are in `test/dattri/algorithm/test_tracin.py`.

dattri/scripts/cifar2_resnet9_benchmark.py
```diff
@@ -0,0 +1,371 @@
...
+            )
+            attributor.cache(train_loader)
+            torch.cuda.reset_peak_memory_stats("cuda")
+            with torch.no_grad():
+                score = attributor.attribute(train_loader, test_loader)
+            peak_memory = torch.cuda.max_memory_allocated("cuda") / 1e6  # Convert to MB
+            print(f"Peak memory usage: {peak_memory} MB")
+
+            target_values = torch.load(
+                "/scratch/bbyo/tli3/cifar2_lds/groundtruth/target_values.pt"
```
TheaperDeng, line 151: local path

dattri/scripts/cifar2_resnet9_benchmark.py
```diff
@@ -0,0 +1,371 @@
...
+            )
+            indices = torch.load(
+                "/scratch/bbyo/tli3/cifar2_lds/groundtruth/training_indices.pt"
```
TheaperDeng, line 154: local path

dattri/scripts/cifar2_resnet9_benchmark.py
```diff
@@ -0,0 +1,371 @@
...
+            print("proj_dim, ensemble", proj_dim, ensemble)
+            projector_kwargs = {
+                "proj_dim": proj_dim,
+                "device": "cuda",
+            }
+            params = []
+            
+            # collect retrained models' paths
+            for i in range(ensemble):
+                params.append(f"/home/shared/dattri-dataset/cifar2_lds_test/models/{i}/model_weights_0.pt")
```
TheaperDeng, line 193: local path

dattri/scripts/cifar2_resnet9_benchmark.py
```diff
@@ -0,0 +1,371 @@
...
+            )
+            attributor.cache(train_loader)
+            torch.cuda.reset_peak_memory_stats("cuda")
+            with torch.no_grad():
+                score = attributor.attribute(test_loader)
+            peak_memory = torch.cuda.max_memory_allocated("cuda") / 1e6  # Convert to MB
+            print(f"Peak memory usage: {peak_memory} MB")
+
+            # get retrained models' location
+            retrain_dir = "/home/shared/dattri-dataset/cifar2_lds_test/models"
```
TheaperDeng, line 211: local path

dattri/scripts/cifar2_resnet9_benchmark.py
```diff
@@ -0,0 +1,371 @@
...
+        # create model
+        model = create_resnet9_model()
+        model.to("cuda")
+
+        for ensemble, normalized_grad in [(1, False), (10, False)]:
+            print("ensemble, normalize grad: ", ensemble, normalized_grad)
+            ensemble = ensemble
+            params = []
+            for i in range(ensemble):
+                params.append(f"/home/shared/dattri-dataset/cifar2_lds_test/models/{i}/model_weights_0.pt")
```
TheaperDeng, line 245: local path

dattri/scripts/cifar2_resnet9_benchmark.py
```diff
@@ -0,0 +1,371 @@
...
+                device=torch.device("cuda"),
+            )
+            torch.cuda.reset_peak_memory_stats("cuda")
+            with torch.no_grad():
+                score = attributor.attribute(train_loader, test_loader)
+            peak_memory = torch.cuda.max_memory_allocated("cuda") / 1e6  # Convert to MB
+            print(f"Peak memory usage: {peak_memory} MB")
+
+            # get retrained models' location
+            retrain_dir = "/home/shared/dattri-dataset/cifar2_lds_test/models"
```
TheaperDeng, line 270: local path

dattri/scripts/cifar2_resnet9_benchmark.py
```diff
@@ -0,0 +1,371 @@
...
+            print("complete\n")
+            # print(args.method, "RESULT:", best_config, "lds:", best_result)
+
+    if args.method == "RPS":
+        model = train_cifar2_resnet9(train_loader, device="cuda")
+        model.eval()
+        model.to("cuda")
+        model.load_state_dict(
+            torch.load(
+                f"/scratch/bbyo/tli3/cifar2-checkpoints/checkpoint_without_dropout/checkpoint_0.pt"
```
TheaperDeng, line 301: local path

dattri/scripts/cifar2_resnet9_benchmark.py
```diff
@@ -0,0 +1,371 @@
...
+            print(f"Peak memory usage: {peak_memory} MB")
+            end_attribute = time.time()
+            print("Attribution time: ", end_attribute - start_attribute)
+            # print(score.shape)
+
+            index_to_value_dict = {value: idx for idx, value in enumerate(all_index)}
+            indices = []
+            # for i in range(50,100):
+            for i in range(50):
+                file_name = f"/scratch/bbyo/tli3/cifar2-checkpoints/checkpoint_without_dropout/selected_indices_seed_{i}.txt"
```
TheaperDeng, line 328: local path

dattri/scripts/cifar2_resnet9_benchmark.py
```diff
@@ -0,0 +1,371 @@
...
+                    # print("indice: ", indice)
+                    indices.append(indice)
+            indices = torch.stack(indices)
+            # print("indices shape: ", indices.shape)
+            print("indices: ", indices)
+
+            target_values = []
+            # for i in range(50,100):
+            for i in range(50):
+                file_name = f"/scratch/bbyo/tli3/cifar2-checkpoints/checkpoint_without_dropout/model_output_checkpoint_{i}.pt"
```
TheaperDeng, line 344: local path

TheaperDeng:
others LGTM

jiaqima (approved):

jiaqima:
@tingwl0122 please feel free to merge this PR when you think it's ready.

PR #98, opened 2024-07-06
[dattri.algorithm] Re-add RPS attributor

### LINKED ISSUE #57: [dattri.algorithm] A few notes and discussions needed of RPS implmentation
Here is a placeholder to describe a few notes on RPS and see people's opinions about my current thoughts.
I will soon raise a PR about the RPS algorithm.
(Update: the PR is created at #58)

To lay some background and for preview, this is my intended `__init__` for the `RPSAttributor`:
```python
class RPSAttributor(BaseAttributor):
    """Representer point selection attributor."""

    def __init__(
        self,
        target_func: Callable,
        model: torch.nn.Module,
        intermediate_layer_name: str,
        final_linear_layer_name: str,
        l2_strength: float = 0.003,
        device: str = "cpu",
    ) -> None:
```

A few notes as follows:
1. **Input** [`target_func`]: As mentioned in #55, RPS doesn't need the gradient of loss w.r.t. params like IF/TracIn/TRAK. But indeed, RPS still needs the gradient of loss w.r.t. "model pre-activation output". I currently still preserve this `target_func` as an input of `RPSAttributor`, but it will serve the purpose of **loss function**. For example, current working test cases use
```python
def f(pre_activation_list, label_list):
    return F.cross_entropy(pre_activation_list, label_list)

```
and 
```python
def f(pre_activation_list, label_list):
    return F.binary_cross_entropy_with_logits(pre_activation_list, label_list)

```
2. **Input** [`intermediate_layer_name`,  `final_linear_layer_name`]:  RPS works on the assumption that the model can be decomposed as a feature model and a prediction model. Basically, it assumes the final layer is something like `nn.Linear(final_feature_dim, class_number)`. As described above, I now allow users to type in the name of the second-to-last layer and the final layer to allow maximum generalizability. 

However, people must hack into the model param space, which is definitely sub-optimal. Another alternative is:
Just enforce this assumption and automatically select these two layers using `model.named_children()`. We will give an error if the final layer is not the type of `nn.Linear`. 

3. **Restriction on loss function and activation**: As described in the paper, their derivation is only natural to **sigmoid+binary cross-entropy loss** for binary classification problems and **softmax+cross-entropy loss** for multi-class classification problems. Their source repo only considers the binary case. I am unsure whether we need to explicitly write this out in the code. Or at least give some warnings if otherwise.

4. **Validity of RPSAttributor.attribute() for multi-class classification**: As derived in the paper, the "representer value for $x_i$ given $x_t$" will be a vector $\in \mathbb{R}^c$, where $c$ stands for the number of label classes. Thus, the attribution score is only valid when solving a **binary classification problem**. (This is probably why RPS and its following work: RPS_LJE only do binary cases in their experiments.) I propose to make this clear for users who want to use RPS or at least raise some warnings.

tingwl0122:
## Description
This PR implements the Representer Point Selection Attributor as another data attribution algorithm.

### 1. Motivation and Context

To migrate the official RPS implementation into our context.
This PR will close #57 .

### 2. Summary of the change

- re-add RPSAttributor and several helper functions to `algorithm/utils`
- add a test file to test RPS functionality
- modify function in `model_utils/hook`
- modify the test function for `model_utils/hook`

### DIFF
--- a/dattri/algorithm/rps.py
+++ b/dattri/algorithm/rps.py
@@ -1,0 +1,181 @@
+"""This module implement the representer point selection."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from typing import Callable
+
+    from torch import Tensor
+    from torch.utils.data import DataLoader
+
+import warnings
+
+import torch
+from torch.nn.functional import normalize
+
+from dattri.algorithm.utils import (
+    _check_shuffle,
+    get_rps_weight,
+    rps_finetune_theta,
+)
+from dattri.model_utils.hook import get_final_layer_io
+
+from .base import BaseAttributor
+
+
+class RPSAttributor(BaseAttributor):
+    """Representer point selection attributor."""
+
+    def __init__(
+        self,
+        loss_func: Callable,
+        model: torch.nn.Module,
+        final_linear_layer_name: str,
+        nomralize_preactivate: bool = False,
+        l2_strength: float = 0.003,
+        epoch: int = 3000,
+        device: str = "cpu",
+    ) -> None:
+        """Representer point selection attributor.
+
+        Args:
+            loss_func (Callable): The loss function to be attributed. The inputs are
+                list of pre-activation values (f_i in the paper) and list of labels.
+                Typical examples are BCELoss and CELoss.
+            model (torch.nn.Module): The model to attribute. RPS will extract
+                second-to-last layer results and the final fc layer's parameter. The
+                second one will be used for the initialization of the l2-finetuning.
+                That is, model output = fc(second-to-last feature).
+            final_linear_layer_name (str): The name of the final linear layer's name
+                in the model.
+            nomralize_preactivate (bool): If set to true, then the intermediate layer
+                output will be normalized. The value of the output inner-product will
+                not affected by the value of individual output magnitude.
+            l2_strength (float): The l2 regularizaton to fine-tune the last layer.
+            epoch (int): The number of epoch used to fine-tune the last layer.
+            device (str): The device to run the attributor. Default is cpu.
+        """
+        self.target_func = loss_func
+        self.model = model
+        self.final_linear_layer_name = final_linear_layer_name
+        self.nomralize_preactivate = nomralize_preactivate
+        self.l2_strength = l2_strength
+        self.epoch = epoch
+        self.device = device
+        self.full_train_dataloader = None
+
+    def cache(self, full_train_dataloader: DataLoader) -> None:
+        """Cache the full dataset for fine-tuning.
+
+        Args:
+            full_train_dataloader (DataLoader): The dataloader
+                with full training samples for the last linear layer
+                fine-tuning.
+        """
+        self.full_train_dataloader = full_train_dataloader
+
+        # get intermediate outputs and predictions for full train dataset
+        intermediate_x_train, y_pred_train = get_final_layer_io(
+            self.model,
+            self.final_linear_layer_name,
+            self.full_train_dataloader,
+            self.device,
+        )
+        # get the initial weight parameter for the final linear layer
+        init_theta = getattr(self.model, self.final_linear_layer_name).weight.data
+        # fine-tuning on the full train dataloader
+        self.finetuned_theta = rps_finetune_theta(
+            intermediate_x_train,
+            y_pred_train,
+            init_theta,
+            loss_func=self.target_func,
+            lambda_l2=self.l2_strength,
+            num_epoch=self.epoch,
+            device=self.device,
+        )
+
+    def attribute(
+        self,
+        train_dataloader: DataLoader,
+        test_dataloader: DataLoader,
+    ) -> Tensor:
+        """Calculate the influence of the training set on the test set.
+
+        Args:
+            train_dataloader (DataLoader): The dataloader for
+                training samples to calculate the influence. It can be a subset
+                of the full training set if `cache` is called before. A subset
+                means that only a part of the training set's influence is calculated.
+                The dataloader should not be shuffled.
+            test_dataloader (DataLoader): The dataloader for
+                test samples to calculate the influence. The dataloader should not\
+                be shuffled.
+
+        Returns:
+            Tensor: The influence of the training set on the test set, with
+                the shape of (num_train_samples, num_test_samples).
+        """
+        super().attribute(train_dataloader, test_dataloader)
+        if self.full_train_dataloader is None:
+            warnings.warn(
+                "The full training data loader was NOT cached. \
+                Treating the train_dataloader as the full training \
+                data loader. And thus the fine-tuned last layer parameters will \
+                also be based on train_dataloader",
+                stacklevel=1,
+            )
+
+        _check_shuffle(train_dataloader)
+        _check_shuffle(test_dataloader)
+
+        intermediate_x_train, y_pred_train = get_final_layer_io(
+            self.model,
+            self.final_linear_layer_name,
+            train_dataloader,
+            device=self.device,
+        )
+
+        # if cache is not called before
+        if self.full_train_dataloader is None:
+            # get the initial weight parameter for the final linear layer
+            init_theta = getattr(self.model, self.final_linear_layer_name).weight.data
+            # finetune the last layer using train samples
+            self.finetuned_theta = rps_finetune_theta(
+                intermediate_x_train,
+                y_pred_train,
+                init_theta,
+                loss_func=self.target_func,
+                lambda_l2=self.l2_strength,
+                num_epoch=self.epoch,
+                device=self.device,
+            )
+
+        # get intermediate features for test samples
+        intermediate_x_test, _ = get_final_layer_io(
+            self.model,
+            self.final_linear_layer_name,
+            test_dataloader,
+            device=self.device,
+        )
+
+        y_test = torch.cat([target for _, target in test_dataloader], dim=0)
+
+        alpha = get_rps_weight(
+            self.finetuned_theta,
+            self.target_func,
+            intermediate_x_train,
+            y_pred_train,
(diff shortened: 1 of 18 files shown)


jiaqima:
Also update dattri/docs/source/api/algorithm.rst?

tingwl0122:
> Also update dattri/docs/source/api/algorithm.rst?

updated.

dattri/algorithm/rps.py
```diff
@@ -0,0 +1,181 @@
...
+
+        Args:
+            full_train_dataloader (DataLoader): The dataloader
+                with full training samples for the last linear layer
+                fine-tuning.
+        """
+        self.full_train_dataloader = full_train_dataloader
+
+        # get intermediate outputs and predictions for full train dataset
+        intermediate_x_train, y_pred_train = get_final_layer_io(
```
TheaperDeng, line 80: Could this be too large for a large full_train_dataloader? Since we will have all the feature and output concanated together as 2 tensors.

dattri/scripts/cifar2_resnet9_benchmark.py
```diff
@@ -0,0 +1,365 @@
...
+            projector_kwargs = {
+                "proj_dim": proj_dim,
+                "device": "cuda",
+            }
+            params = []
+
+            # collect retrained models' paths
+            for i in range(ensemble):
+                params.append(
+                    f"/home/shared/dattri-dataset/cifar2_lds_test/models/{i}/model_weights_0.pt"
```
TheaperDeng, line 198: local path leak

dattri/scripts/cifar2_resnet9_benchmark.py
```diff
@@ -0,0 +1,365 @@
...
+        model = create_resnet9_model()
+        model.to("cuda")
+
+        for ensemble, normalized_grad in [(1, False), (10, False)]:
+            print("ensemble, normalize grad: ", ensemble, normalized_grad)
+            ensemble = ensemble
+            params = []
+            for i in range(ensemble):
+                params.append(
+                    f"/home/shared/dattri-dataset/cifar2_lds_test/models/{i}/model_weights_0.pt"
```
TheaperDeng, line 254: local path leak

dattri/scripts/cifar2_resnet9_benchmark.py
```diff
@@ -0,0 +1,365 @@
...
+                device=torch.device("cuda"),
+            )
+            torch.cuda.reset_peak_memory_stats("cuda")
+            with torch.no_grad():
+                score = attributor.attribute(train_loader, test_loader)
+            peak_memory = torch.cuda.max_memory_allocated("cuda") / 1e6  # Convert to MB
+            print(f"Peak memory usage: {peak_memory} MB")
+
+            # get retrained models' location
+            retrain_dir = "/home/shared/dattri-dataset/cifar2_lds_test/models"
```
TheaperDeng, line 280: local path leak

dattri/scripts/cifar2_resnet9_benchmark.py
```diff
@@ -0,0 +1,365 @@
...
+            print("complete\n")
+            # print(args.method, "RESULT:", best_config, "lds:", best_result)
+
+    if args.method == "RPS":
+        model = create_resnet9_model()
+        model.eval()
+        model.to("cuda")
+        model.load_state_dict(
+            torch.load(
+                f"/home/shared/dattri-dataset/cifar2_lds_test/models/0/model_weights_0.pt"
```
TheaperDeng, line 313: local path leak

dattri/scripts/cifar2_resnet9_benchmark.py
```diff
@@ -0,0 +1,365 @@
...
+                score = attributor.attribute(train_loader, test_loader)
+                peak_memory = (
+                    torch.cuda.max_memory_allocated("cuda") / 1e6
+                )  # Convert to MB
+                print(f"Peak memory usage: {peak_memory} MB")
+                end_attribute = time.time()
+                print("Attribution time: ", end_attribute - start_attribute)
+
+                # get retrained models' location
+                retrain_dir = "/home/shared/dattri-dataset/cifar2_lds_test/models"
```
TheaperDeng, line 339: local path leak

dattri/scripts/cifar2_resnet9_benchmark.py
```diff
@@ -0,0 +1,365 @@
...
+            with torch.no_grad():
+                score = attributor.attribute(train_loader, test_loader)
+            peak_memory = torch.cuda.max_memory_allocated("cuda") / 1e6  # Convert to MB
+            print(f"Peak memory usage: {peak_memory} MB")
+
+            target_values = torch.load(
+                "/scratch/bbyo/tli3/cifar2_lds/groundtruth/target_values.pt"
+            )
+            indices = torch.load(
+                "/scratch/bbyo/tli3/cifar2_lds/groundtruth/training_indices.pt"
```
TheaperDeng, line 158: local path leak

dattri/scripts/cifar2_resnet9_benchmark.py
```diff
@@ -0,0 +1,365 @@
...
+            )
+            attributor.cache(train_loader)
+            torch.cuda.reset_peak_memory_stats("cuda")
+            with torch.no_grad():
+                score = attributor.attribute(train_loader, test_loader)
+            peak_memory = torch.cuda.max_memory_allocated("cuda") / 1e6  # Convert to MB
+            print(f"Peak memory usage: {peak_memory} MB")
+
+            target_values = torch.load(
+                "/scratch/bbyo/tli3/cifar2_lds/groundtruth/target_values.pt"
```
TheaperDeng, line 155: local path leak

test/dattri/algorithm/test_rps.py
```diff
@@ -0,0 +1,117 @@
...
+        test_loader = DataLoader(dataset, batch_size=2)
+        train_loader = DataLoader(
+            dataset,
+            batch_size=2,
+            sampler=SubsetSampler(range(4)),
+        )
+
+        # define the loss function
+        def f(pre_activation_list, label_list):
+            return functional.cross_entropy(pre_activation_list, label_list)
```
TheaperDeng, line 55: so we have a somehow different target function than IF/TRAK/TracIN, but I think it's fine.

test/dattri/algorithm/test_rps.py
```diff
@@ -0,0 +1,117 @@
...
+
+        # define the RPS attributor w/ cache
+        attributor = RPSAttributor(
+            loss_func=f,
+            model=model,
+            final_linear_layer_name="fc3",
```
TheaperDeng, line 61: this layer name is kind of different from `list(model.named_paramters()).keys()` right? I guess the keys might be `fc3.weights` and `fc3.bias` for this layer. We may need a better explanation in the docuemnt.

dattri/algorithm/rps.py
```diff
@@ -0,0 +1,181 @@
...
+        l2_strength: float = 0.003,
+        epoch: int = 3000,
+        device: str = "cpu",
+    ) -> None:
+        """Representer point selection attributor.
+
+        Args:
+            loss_func (Callable): The loss function to be attributed. The inputs are
+                list of pre-activation values (f_i in the paper) and list of labels.
+                Typical examples are BCELoss and CELoss.
```
tingwl0122, line 46: Still Called it a `target_func` but add some additional comments to state this is typically a loss function.

dattri/model_utils/hook.py
```diff
@@ -8,30 +9,35 @@
...
     from typing import List, Tuple
 
+    from torch.utils.data import DataLoader
+
 import torch
 from torch import Tensor
 
 
-def get_layer_feature(
+def _get_layer_feature(
```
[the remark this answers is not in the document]
tingwl0122: Sorry for the confusion. Since now I assume that the user must input a model with a final linear layer, the first one will be deprecated. (The second one assumes the model to have a final linear layer and directly get the input/output of that layer, which will be inputs of RPS)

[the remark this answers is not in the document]
tingwl0122: Will remove `_get_layer_feature` and add some words in the docstring of `get_final_layer_io` and `RPSAttributor`.

test/dattri/algorithm/test_rps.py
tingwl0122: will still name it as `target_func` for now

dattri/algorithm/rps.py
tingwl0122: Yeah, you are right. If this is really an issue, we can probably do this in a batch manner?

dattri/scripts/cifar2_resnet9_benchmark.py
```diff
@@ -0,0 +1,365 @@
+"""This script shows how to use dattri algorithms for benchmarking CIFAR-2."""
```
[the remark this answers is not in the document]
tingwl0122: this should be convenient for users to run benchmark experiments using their own data/model?
now it is customized for cifar2 but it can be generalized to different exps just like the retrain script.
What is the reason you want to remove this?

test/dattri/algorithm/test_rps.py
tingwl0122: Yeah, now it is the "layer name" but not "parameter name". Now for the fine-tuning step, I only choose to do it on the weight matrix (ignoring the bias entirely). If this is really a matter, we can do the "add-1-dim" trick to incorporate the bias into the weight matrix first then do the fine-tuning.

dattri/algorithm/rps.py
tingwl0122: Now I think this should be minor? The peak mem of RPS is always much smaller than all the other methods.

TheaperDeng:
LGTM.

TheaperDeng (approved):

PR #99, opened 2024-07-08
[dattri.algorithm] Add `AttributionTask` abstraction for new API design

TheaperDeng:
### 1. Motivation and Context
1. Introduce a new `AttributionTask` class to manage model and checkpoint information, in order to reduce user's workload on extracting internal information of the models needed for some algorithms.
2. Make 2 levels of Attributor API, which makes it easier for users to build customized attributors from our template rather than directly from scratch.
3. Support DataInf as one of the attributors.
4. All examples, unittests and API documents are changed accordingly.

### 2. Summary of the change
1. Introduce a new `AttributionTask` API, so that users will follow the following template for attributor usage
```python
# User code
model = create_model()

def f(params, data_target_pair):
    image, label = data_target_pair
    loss = nn.CrossEntropyLoss()
    yhat = torch.func.functional_call(model, params, image)
    return loss(yhat, label.long())

task = AttributionTask(target_func=f, model=model, checkpoints=model.state_dict())
attributor = Attributor(task=task, device="cuda")
```

This `AttributionTask` contains some useful member functions that are very helpful for attributor development.
```python
# Developer code
task.get_grad_target_func(...)  # get the targeted function's gradient func
task.get_target_func(...)  # get the targeted function, which can be directly used by dattri.func
task.get_param(...)  # get the parameters w/ or w/o layer split, as well as the layer mapping automatically (for datainf and ek-fac)
task.register_forward_hook(...)  # TODO, return the handler of the hook for hidden states.
```

2. Refactors the original IFAttributors into a more flexible `BaseInnerProductAttributor` and some high-level implementations of Attributors.

Users who do not want to customize their own attributor may directly use the high-level implementations of Attributors.

```python
# User code
from dattri.algorithm.influence_function import\
    IFAttributorExplicit,\
    IFAttributorCG,\
    IFAttributorLiSSA,\
    IFAttributorDataInf,\
    IFAttributorArnoldi

attributor = IFAttributorExplicit(task=task, device="cuda", regularization=0.01)
attributor = IFAttributorCG(task=task, device="cuda", regularization=0.01)
attributor = IFAttributorLiSSA(task=task, device="cuda", recursion_depth=100)
attributor = IFAttributorDataInf(task=task, device="cuda", regularization=0.01)
attributor = IFAttributorArnoldi(task=task, device="cuda", regularization=0.01)

# following usage is not changed (i.e., attributor.cache() and attributor.attribute())
```

Developers who want to create their own attributors may follow the templates we provided, currently we have `BaseInnerProductAttributor` for a large family of gradient-based TDA methods that has (g^TH^{-1}g) format.

It can also support
- GradCos/GradDot
- IF
- Embedding similarity based TDA methods

```python
class MyAttributor(BaseInnerProductAttributor):
    def generate_test_query(
        self,
        index: int,
        data: Tuple[torch.Tensor, ...],
    ) -> torch.Tensor:
        # this methods defaultly calculate the gradient over test samples
        # but developers may want to have their own design, .e.g, use some projectors.
        model_params, _ = self.task.get_param(index)
        return self.task.get_grad_target_func()(model_params, data)

    def generate_train_query(
        self,
        index: int,
        data: Tuple[torch.Tensor, ...],
    ) -> torch.Tensor:
        # this methods defaultly calculate the gradient over train samples
        # but developers may want to have their own design, .e.g, use some projectors.
        model_params, _ = self.task.get_param(index)
        return self.task.get_grad_target_func()(model_params, data)

    def transformation_on_query(
        self,
        index: int,
        data: Tuple[torch.Tensor, ...],
        query: torch.Tensor,
        **transformation_kwargs
    ) -> torch.Tensor:
       # This methods is called on the test query (as the vector)
       # IF will calculate the Hessian/Fisher and make the ihvp or ifvp
       # Grad-cos/dot will directly return the test query.

       # Let's take explicitly IHVP as example
       from dattri.func.ihvp import ihvp_explicit

        self.ihvp_func = ihvp_explicit(
            partial(self.task.get_target_func(), data_target_pair=data),
            **transformation_kwargs,
        )
        model_params, _ = self.task.get_param(index)
        return self.ihvp_func((model_params,), query).detach()
```

3. DataInf can be directly supported under this design, **with no additional work on the users' side**.

- [x] Document test: If you added an external API, then you should check if the document is correctly generated.

### 4. TODOs
Some todos will be added to the project tracking page.

### DIFF
--- a/dattri/algorithm/influence_function.py
+++ b/dattri/algorithm/influence_function.py
@@ -5,99 +5,142 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    from typing import Any, Callable, Dict, List, Optional, Tuple
+    from typing import Any, Dict, List, Optional, Tuple
 
     from torch import Tensor
     from torch.utils.data import DataLoader
 
+    from dattri.task import AttributionTask
+
 import warnings
+from abc import abstractmethod
 from functools import partial
 
 import torch
-from torch.func import grad, vmap
 from tqdm import tqdm
-
-from dattri.func.ihvp import ihvp_arnoldi, ihvp_cg, ihvp_explicit, ihvp_lissa
-from dattri.func.utils import flatten_params
 
 from .base import BaseAttributor
 from .utils import _check_shuffle
 
 
-def _lissa_collate_fn(
-    sampled_input: List[Tensor],
-) -> Tuple[Tensor, List[Tuple[Tensor, ...]]]:
-    """Collate function for LISSA.
-
-    Args:
-        sampled_input (List[Tensor]): The sampled input from the dataloader.
-
-    Returns:
-        Tuple[Tensor, List[Tuple[Tensor, ...]]]: The collated input for the LISSA.
-    """
-    return (
-        sampled_input[0],
-        tuple(sampled_input[i].float() for i in range(1, len(sampled_input))),
-    )
-
-
-SUPPORTED_IHVP_SOLVER = {
-    "explicit": ihvp_explicit,
-    "cg": ihvp_cg,
-    "arnoldi": ihvp_arnoldi,
-    "lissa": partial(ihvp_lissa, collate_fn=_lissa_collate_fn),
-}
-
-
-class IFAttributor(BaseAttributor):
-    """Influence function attributor."""
+class BaseInnerProductAttributor(BaseAttributor):
+    """The base class for inner product attributor."""
 
     def __init__(
         self,
-        target_func: Callable,
-        params: dict,
-        ihvp_solver: str = "explicit",
-        ihvp_kwargs: Optional[Dict[str, Any]] = None,
-        device: str = "cpu",
+        task: AttributionTask,
+        device: Optional[str] = "cpu",
+        **transformation_kwargs: Dict[str, Any],
     ) -> None:
-        """Influence function attributor.
+        """Initialize the attributor.
 
         Args:
-            target_func (Callable): The target function to be attributed.
-                The function can be quite flexible in terms of what is calculate,
-                but it should take the parameters and the dataloader as input.
-                A typical example is as follows:
-                ```python
-                @flatten_func(model)
-                def f(params, dataloader):
-                    loss = nn.CrossEntropyLoss()
-                    loss_val = 0
-                    for image, label in dataloader:
-                        yhat = torch.func.functional_call(model, params, image)
-                        loss_val += loss(yhat, label)
-                    return loss_val
-                ```.
-                This examples calculates the loss of the model on the dataloader.
-            params (dict): The parameters of the target function. The key is the
-                name of a parameter and the value is the parameter tensor.
-                TODO: This should be changed to support a list of parameters or
-                    paths for ensembling and memory efficiency.
-            ihvp_solver (str): The solver for inverse hessian vector product
-                calculation, currently we only support "explicit", "cg" and "arnoldi".
-            ihvp_kwargs (Optional[Dict[str, Any]]): Keyword arguments for ihvp solver.
-                calculation, currently we only support "explicit", "cg", "arnoldi",
-                and "lissa".
-            device (str): The device to run the attributor. Default is cpu.
+            task (AttributionTask): The task to be attributed. The task should
+                be an instance of `AttributionTask`.
+            transformation_kwargs (Optional[Dict[str, Any]]): The keyword arguments for
+                the transformation function.
+            device (str): The device to run the attributor.
         """
-        self.target_func = target_func
-        self.params = flatten_params(params)
-        if ihvp_kwargs is None:
-            ihvp_kwargs = {}
-        self.ihvp_solver_name = ihvp_solver
-        self.ihvp_solver = partial(SUPPORTED_IHVP_SOLVER[ihvp_solver], **ihvp_kwargs)
+        self.task = task
+        self.transformation_kwargs = transformation_kwargs or {}
         self.device = device
-        self.full_train_dataloader = None
-        self.grad_func = vmap(grad(self.target_func), in_dims=(None, 1))
+        self.index = 0
+
+    def _set_test_data(self, dataloader: torch.utils.data.DataLoader) -> None:
+        """Set test dataloader."""
+        # This function may be overrided by the subclass
+        self.test_dataloader = dataloader
+
+    def _set_train_data(self, dataloader: torch.utils.data.DataLoader) -> None:
+        """Set train dataloader to be attributed."""
+        # This function may be overrided by the subclass
+        self.train_dataloader = dataloader
+
+    def _set_full_train_data(self, dataloader: torch.utils.data.DataLoader) -> None:
+        """Set train dataloader for full training set."""
+        # This function may be overrided by the subclass
+        self.full_train_dataloader = dataloader
+
+    def generate_test_query(
+        self,
+        index: int,
+        data: Tuple[torch.Tensor, ...],
+    ) -> torch.Tensor:
+        """Calculating the query based on the test data.
+
+        Args:
+            index (int): The index of the model parameters. This index
+                is used for ensembling of different trained model
+                parameters.
+            data (Tuple[Tensor]): The test data. Normally this is a tuple
+                of input data and target data, the number of items in the
+                tuple should be aligned in the target function. The tensors'
+                shape follows (1, batchsize, ...).
+
+        Returns:
+            torch.Tensor: The query based on the test data. Normally it is
+                a 2-d dimensional tensor with the shape of
+                (batchsize, num_parameters).
+        """
+        model_params, _ = self.task.get_param(index)
+        return self.task.get_grad_target_func()(model_params, data)
+
+    def generate_train_query(
+        self,
+        index: int,
+        data: Tuple[torch.Tensor, ...],
+    ) -> torch.Tensor:
+        """Calculating the query based on the train data.
+
+        Args:
+            index (int): The index of the model parameters. This index
+                is used for ensembling of different trained model
+                parameters.
+            data (Tuple[Tensor]): The train data. Normally this is a tuple
+                of input data and target data, the number of items in the
+                tuple should be aligned in the target function. The tensors'
+                shape follows (1, batchsize, ...).
+
+        Returns:
+            torch.Tensor: The query based on the train data. Normally it is
+                a 2-d dimensional tensor with the shape of
+                (batchsize, num_parameters).
+        """
+        model_params, _ = self.task.get_param(index)
+        return self.task.get_grad_target_func()(model_params, data)
+
+    @abstractmethod
+    def transformation_on_query(
+        self,
+        index: int,
+        data: Tuple[torch.Tensor, ...],
+        query: torch.Tensor,
+    ) -> torch.Tensor:
+        """Calculate the transformation on the query.
+
+        Inner product attributor calculates the inner product between the
+        test query and the transformation of the test query. This function
+        calculates the transformation of the test query based on the train data.
+
+        Normally speaking, this function may return any transformation on the query.
+        Hessian Matrix and Fisher Information Matrix are two common transformations.
+
+        Args:
+            index (int): The index of the model parameters. This index
+                is used for ensembling of different trained model.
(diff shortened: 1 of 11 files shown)


dattri/func/ihvp.py
```diff
@@ -877,29 +893,38 @@ def ihvp_at_x_datainf(
...
     """
+
     def _per_sample_grad(*args) -> torch.Tensor:
         return grad(func, argnums=argnums)(*args)
 
-    def _single_datainf_ihvp(v: torch.Tensor,
-                                     grad: torch.Tensor,
-                                     regularization: float) -> torch.Tensor:
-        coef = (v.T @ grad) / (regularization + torch.sum(grad ** 2))
+    def _single_datainf_ihvp(
```
tingwl0122, line 900: It seems like there is a redundancy for this function and the above `_ihvp_datainf_func`?
These codes look really similar.

dattri/func/ihvp.py
```diff
@@ -877,29 +893,38 @@ def ihvp_at_x_datainf(
...
+                    if param_layer_map[layer] == group
+                ],
+            )
             concated_grads = torch.concat(grouped_layers, dim=1)
             grouped.append(concated_grads)
         grads = tuple(grouped)
         layer_cnt = max_layer + 1  # Assuming count starts from 0
 
-    def _ihvp_at_x_datainf_func(v: Tuple[torch.Tensor, ...],
+    def _ihvp_at_x_datainf_func(
```
tingwl0122, line 926: And also this function and the last part of `_ihvp_datainf_func`.
It is possible to optimize a bit?

dattri/func/ihvp.py
```diff
@@ -1225,11 +1254,13 @@ def cached_forward(self: torch.nn.Module, *args, **kwrds) -> torch.Tensor:
         cache.check_type()
         cache.retain_grad()
         return outputs
+
     return cached_forward
 
 
 class MLPCache:
     """Cache of input and output variables in a MLP layer."""
```
tingwl0122, line 1262: I am thinking whether we can just use forward_hook for this part.
I know this will probably not be addressed by this PR, but just for simplicity.

dattri/algorithm/influence_function.py
```diff
@@ -5,99 +5,154 @@
...
-        Tuple[Tensor, List[Tuple[Tensor, ...]]]: The collated input for the LISSA.
-    """
-    return (
-        sampled_input[0],
-        tuple(sampled_input[i].float() for i in range(1, len(sampled_input))),
-    )
+        Args:
+            task (AttributionTask): The task to be attributed. The task should
+                be an instance of `AttributionTask`.
+            transformation_kwargs (Optional[Dict[str, Any]]): The keyword arguments for
```
tingwl0122, line 40: I think we can be more informative for this term: "transformation"?

dattri/algorithm/influence_function.py
```diff
@@ -5,99 +5,154 @@
...
-        params: dict,
-        ihvp_solver: str = "explicit",
-        ihvp_kwargs: Optional[Dict[str, Any]] = None,
-        device: str = "cpu",
-    ) -> None:
-        """Influence function attributor.
+        index: int,
+        data: Tuple[torch.Tensor, ...],
+    ) -> torch.Tensor:
+        """Calculating the query based on the test data.
```
tingwl0122, line 81: quite vague for me of the term: "query". I don't think this is a standard terminology in the context of IF?
Should we borrow some intro from `transformation_on_query`?

dattri/algorithm/influence_function.py
```diff
@@ -5,99 +5,154 @@
...
+    def transformation_on_query(
+        self,
+        index: int,
+        data: Tuple[torch.Tensor, ...],
+        query: torch.Tensor,
+    ) -> torch.Tensor:
+        """Calculate the transformation on the query.
+
+        Inner product attributor calculates the inner product between the
+        test query and the transformation of the test query. This function
```
tingwl0122, line 134: between test query and the transformatino of the "train" query? or probably the opposite?

dattri/algorithm/influence_function.py
```diff
@@ -5,99 +5,154 @@
...
+        calculates the transformation of the test query based on the train data.
+
+        Normally speaking, this function may return any transformation on the query.
+        Hessian Matrix and Fisher Information Matrix are two common transformations.
```
tingwl0122, line 138: We can also mention this for the first time the term: "transformation" appears.

dattri/algorithm/influence_function.py
```diff
@@ -5,99 +5,154 @@
...
+        self,
+        index: int,
+        data: Tuple[torch.Tensor, ...],
+        query: torch.Tensor,
+    ) -> torch.Tensor:
+        """Calculate the transformation on the query.
+
+        Inner product attributor calculates the inner product between the
+        test query and the transformation of the test query. This function
+        calculates the transformation of the test query based on the train data.
```
tingwl0122, line 135: transformation of the "train" query based on the train data?
but do we care about what kind of data we used to call this function?

dattri/algorithm/influence_function.py
```diff
@@ -5,99 +5,154 @@
...
+
+        Normally speaking, this function may return any transformation on the query.
+        Hessian Matrix and Fisher Information Matrix are two common transformations.
+
+        Args:
+            index (int): The index of the model parameters. This index
+                is used for ensembling of different trained model.
+            data (Tuple[Tensor]): The train data. Normally this is a tuple
```
tingwl0122, line 143: if this is always train data, then just call it train_data?

dattri/algorithm/influence_function.py
```diff
@@ -5,99 +5,154 @@
...
+                of input data and target data, the number of items in the
+                tuple should be aligned in the target function. The tensors'
+                shape follows (1, batchsize, ...).
```
tingwl0122, line 146: Is this because we (nearly always) use `unsqueeze(0)` on the data before?

dattri/algorithm/influence_function.py
```diff
@@ -5,99 +5,154 @@
...
+            query (torch.Tensor): The query based on the test data. Normally it is
```
tingwl0122, line 147: if this is always associated with test data, then just call it test_query?

tingwl0122 (commented):
Generally the structure of `dattri/func/task` looks good to me. Left minor comments for readability and clarity of the functions and docstrings.

dattri/algorithm/influence_function.py
TheaperDeng: I guess this one is kind of getting more and more popular in the TDA community.

I will add some intro from `transformation_on_query` here.

dattri/algorithm/influence_function.py
TheaperDeng: I am afraid so, yes.

dattri/algorithm/influence_function.py
TheaperDeng: I will change the comment to "The query to be transformed."

dattri/func/ihvp.py
```diff
@@ -784,16 +785,23 @@ def ihvp_datainf(
         `v`(layer-wise) and returns the approximated IHVP of the approximated Hessian of
         `func` and `v`.
     """
+    # TODO: param_layer_map should not be optional.
```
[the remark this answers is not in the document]
TheaperDeng: The reason why `param_layer_map` will always have input if users use `Attributor`. But it should not be `None` if users use the `ihvp_datainf` directly (currently there is no guarantee). So I leave a TODO here since it's a minor problem.

dattri/func/ihvp.py
```diff
@@ -784,16 +785,23 @@ def ihvp_datainf(
...
-    def _single_datainf_ihvp(v: torch.Tensor,
-                             grad: torch.Tensor,
-                             regularization: float) -> torch.Tensor:
-        coef = (v.T @ grad) / (regularization + torch.sum(grad ** 2))
-        return (v - coef * grad) / regularization
+    def _single_datainf_ihvp(
+        v: torch.Tensor,
+        grad: torch.Tensor,
+        regularization: float,
+    ) -> torch.Tensor:
```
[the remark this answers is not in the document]
TheaperDeng: will add a todo for @charles-pyj

dattri/func/ihvp.py
TheaperDeng: will add a todo for @charles-pyj

dattri/func/ihvp.py
TheaperDeng: leave a todo for @charles-pyj

dattri/func/ihvp.py
TheaperDeng: Yes that's our plan, will do for next PR. Had some discussion with @sx-liu already.

dattri/task.py
```diff
@@ -0,0 +1,238 @@
...
+
+import torch
+from torch import nn
+from torch.func import grad, vmap
+
+from dattri.func.utils import flatten_func, flatten_params
+
+
+class AttributionTask:
+    """The abstraction of the abstraction task."""
```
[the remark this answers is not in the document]
TheaperDeng: fix a little, but it only allow me to write 1 line (<90 char), I can not make it better :)

dattri/task.py
```diff
@@ -0,0 +1,238 @@
...
+            self.checkpoints = [checkpoints]
+        else:
+            self.checkpoints = checkpoints
+
+        # current_checkpoint_idx is used to state
+        # which checkpoint is currently loaded.
+        self.current_checkpoint_idx = None
+
+        # TODO: Make this more general, that is allow customized kwargs.
+        self.grad_func = vmap(grad(self.target_func), in_dims=(None, 1))
```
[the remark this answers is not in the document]
TheaperDeng: I think this is written by you in previous PR. That's because of the `unsqueeze(0)` we used.

dattri/task.py
```diff
@@ -0,0 +1,238 @@
...
+
+    def _load_checkpoints(self, index: int) -> None:
+        """This method load the checkpoint at the specified index.
+
+        Args:
+            index (int): The index of the checkpoint to be loaded.
+        """
+        if self.current_checkpoint_idx is None or self.current_checkpoint_idx != index:
```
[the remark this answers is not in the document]
TheaperDeng: Yes, it's a todo, the task will support the ensemble as well (but in future PRs).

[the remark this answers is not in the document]
TheaperDeng: No, it's just used as a record for which checkpoint is loaded in the model.

dattri/task.py
```diff
@@ -0,0 +1,238 @@
...
+            else:
+                self.model.load_state_dict(self.checkpoints[index])
+            self.current_checkpoint_idx = index
+            self.named_parameters = {
+                k: p for k, p in self.model.named_parameters() if p.requires_grad
+            }
+
+    @staticmethod
+    def _genearte_param_layer_map(
+        named_parameters: Dict[str, torch.Tensor],
```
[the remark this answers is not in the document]
TheaperDeng: No I think we can't, or we need some other parameters in this function. Because users may just want to carry out TDA methods on a portion of the parameters (e.g., some specific layers), so named_parameters may only contain a portion of layers.

dattri/task.py
```diff
@@ -0,0 +1,238 @@
...
+
+        Raises:
+            NotImplementedError: If layer_name is not None.
+        """
+        if layer_name is not None:
+            error_msg = "layer_name has not been implemented yet."
+            raise NotImplementedError(error_msg)
+        return self.grad_func
+
+    def get_target_func(
```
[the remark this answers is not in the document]
TheaperDeng: For inverse hessian and fisher information calculation.

dattri/task.py
```diff
@@ -0,0 +1,238 @@
...
+
+        Raises:
+            NotImplementedError: If layer_name is not None.
+        """
+        if layer_name is not None:
+            error_msg = "layer_name has not been implemented yet."
+            raise NotImplementedError(error_msg)
+        return self.target_func
+
+    def get_param(
```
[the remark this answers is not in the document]
TheaperDeng: No it's generally for all TDA algorithms. since TDA developers will not have touch to the model anymore, they rely on the api provided by the `task` to get the parameters.

PR #100, opened 2024-07-12
[dattri.func] Re-define Arnoldi algorithm as a projection method

tingwl0122:
### 1. Motivation and Context
Originally, Arnoldi algorithm is defined as one of the ihvp method in `dattri/func/ihvp.py`, but funcitonal-wise, it is closer to a projection method with approximation on the inverse-Hessian. Thus, we move it to `dattri/func/projection.py`.

### 2. Summary of the change
- Re-name `dattri/func/random_projection.py` as `dattri/func/projection.py`
- Define a new class of projector named `ArnoldiProjector` to take care of the Arnoldi algorithm.


dattri/scripts/cifar2_resnet9_benchmark.py
```diff
@@ -0,0 +1,371 @@
+"""This example shows how to use the IF to detect noisy labels in the MNIST."""
```
TheaperDeng, line 1: maybe remove this file since we have it in our repo?

dattri/scripts/mnist_lr_benchmark.py
```diff
@@ -0,0 +1,379 @@
+"""This example shows how to use the IF to detect noisy labels in the MNIST."""
```
TheaperDeng, line 1: maybe add it to `experiments/`?  See #105

dattri/scripts/mnist_lr_benchmark.py
```diff
@@ -0,0 +1,379 @@
...
+                device=torch.device("cuda"),
+            )
+            torch.cuda.reset_peak_memory_stats("cuda")
+            with torch.no_grad():
+                score = attributor.attribute(train_loader, test_loader)
+            peak_memory = torch.cuda.max_memory_allocated("cuda") / 1e6  # Convert to MB
+            print(f"Peak memory usage: {peak_memory} MB")
+
+            # get retrained models' location
+            retrain_dir = "/home/shared/dattri-dataset/cifar2_lds_test/models"
```
TheaperDeng, line 276: some local paths in this file.

TheaperDeng (commented):
The projector LGTM, so we still need another PR to make this support `IFAttributorArnoldi` right?

tingwl0122:
> The projector LGTM, so we still need another PR to make this support `IFAttributorArnoldi` right?

Yes, or do you think it is better to include that in this PR?

tingwl0122:
> > The projector LGTM, so we still need another PR to make this support `IFAttributorArnoldi` right?
> 
> Yes, or do you think it is better to include that in this PR?

probably not, I will do this in the next PR.

tingwl0122:
Closed since commit history is too messy, a new PR will replace this.

PR #104, opened 2024-07-16
[dattri.algorithm] Support `AttributionTask` API in TracIN/TRAK/RPS Attributors

TheaperDeng:
### 1. Motivation and Context
We have a new API in #99 . We want to make all the attributors to support this API.

**Note: The support of TracIN/TRAK/RPS Attributors aims to involve less user code change.** 

### 2. Summary of the change
1. Support `AttributionTask` API in TracIN/TRAK/RPS Attributors, includes all unittest, examples and documents.
2. Support ensemble for all attributors based on innerproduct


dattri/task.py
```diff
@@ -142,17 +152,23 @@ def get_grad_target_func(
         if layer_name is not None:
             error_msg = "layer_name has not been implemented yet."
             raise NotImplementedError(error_msg)
+
+        if self.grad_func_kwargs != {"in_dims": in_dims}:
+            self.grad_func = vmap(grad(self.target_func), in_dims=in_dims)
```
tingwl0122, line 157: do we test these different `in_dims`?

tingwl0122 (commented):
Overall looks good to me. Does not catch some obvious bugs.
We can probably re-run some experiments/scripts to make sure these algs still have the same functionalities as before

dattri/task.py
TheaperDeng: No, but it's directly called by the attributors, so I think it's fine for now. If the `in_dims` is not correct, an error will be reported by `vmap`.

tingwl0122:
Just raised an issue (#107) to keep track of the benchmark performance using the new API in this PR.

PR #106, opened 2024-07-16
[dattri.benchmark] add benchmark downloading API

TheaperDeng:
### 1. Motivation and Context
We need an API help users to download the groundtruth of some benchmark settings as well as the trained model.

The dataset is hosted on https://huggingface.co/datasets/trais-lab/dattri-benchmark/tree/main

### 2. Summary of the change
```python
from dattri.benchmark.load import load_benchmark
from dattri.task import AttributionTask
from dattri.metrics import lds

model_details, groundtruth = load_benchmark(model="mlp", dataset="mnist", metric="lds")
task = AttributionTask(
    model=model_details["model"],
    target_func=model_details["target_func"],
    checkpoints=model_details["model_full"]
)
attributor = Attributor(task=task)
score = attributor.attribute(
    DataLoader(model_details["train_dataset"], shuffle=False),
    DataLoader(model_details["test_dataset"], shuffle=False),
)
print("lds:" lds(score, groundtruth))
```

Will add examples in later PRs.

### DIFF
--- a/dattri/benchmark/load.py
+++ b/dattri/benchmark/load.py
@@ -175,7 +175,8 @@ def load_benchmark(
                 for the algorithm you want to benchmark.
 
             The second tuple contains the ground truth for the benchmark,
-            the items are listed as following. It can be directly sent to the
+            the items are subjected to change for each benchmark settings.
+            It can be directly sent to the
             metrics function defined in `dattri.metrics`. Notably, the groundtruth
             depends on the `metric` parameter user stated.
 
@@ -227,7 +228,7 @@ def load_benchmark(
         / identifier
         / "models_full"
         / f"{i}"
-        / f"model_weights_{i}.pt"
+        / "model_weights_0.pt"
         for i in range(models_full_count)
     ]
     models_half_count = _count_folders(
@@ -239,7 +240,7 @@ def load_benchmark(
         / identifier
         / "models_half"
         / f"{i}"
-        / f"model_weights_{i}.pt"
+        / "model_weights_0.pt"
         for i in range(models_half_count)
     ]
     train_dataset, test_dataset = SUPPORTED_DATASETS[dataset](download_path / "dataset")
--- a/dattri/benchmark/load.py
+++ b/dattri/benchmark/load.py
@@ -26,6 +26,7 @@ from dattri.benchmark.datasets.mnist import (
     train_mnist_mlp,
 )
 from dattri.benchmark.models import SUPPORTED_MODELS
+from dattri.benchmark.utils import SubsetSampler
 
 REPO_URL = "https://huggingface.co/datasets/trais-lab/dattri-benchmark/resolve/main/"
 
@@ -257,11 +258,13 @@ def load_benchmark(
     )
 
     return {
-        "model": MODEL_MAP[identifier](),
+        "model": MODEL_MAP[identifier]().eval(),
         "models_full": models_full_list,
         "models_half": models_half_list,
         "train_dataset": train_dataset,
         "test_dataset": test_dataset,
+        "train_sampler": SubsetSampler(range(5000)),
+        "test_sampler": SubsetSampler(range(500)),
         "loss_func": loss_func,
         "target_func": target_func,
         "train_func": train_func,
--- a/dattri/benchmark/load.py
+++ b/dattri/benchmark/load.py
@@ -25,7 +25,6 @@ from dattri.benchmark.datasets.mnist import (
     train_mnist_lr,
     train_mnist_mlp,
 )
-from dattri.benchmark.models import SUPPORTED_MODELS
 from dattri.benchmark.utils import SubsetSampler
 
 REPO_URL = "https://huggingface.co/datasets/trais-lab/dattri-benchmark/resolve/main/"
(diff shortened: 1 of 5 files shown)


dattri/benchmark/load.py
```diff
@@ -0,0 +1,268 @@
...
+import os
+import pathlib
+import zipfile
+from functools import partial
+from io import BytesIO
+
+import requests
+import torch
+
+from dattri.benchmark.datasets.mnist import (
```
tingwl0122, line 19: So this is just an example using MNIST or we now only have the stored ckpts/gts for MNIST?

dattri/benchmark/load.py
```diff
@@ -0,0 +1,268 @@
...
+            with (destination_path / file_name).open(mode="wb") as file:
+                for chunk in response.iter_content(chunk_size=1024):
+                    if chunk:
+                        file.write(chunk)
+    else:
+        error_msg = f"Failed to download file. Status code: {response.status_code}"
+        raise ValueError(error_msg)
+
+
+def load_benchmark(
```
tingwl0122, line 128: do we need a test file or script for this function?

dattri/benchmark/load.py
```diff
@@ -0,0 +1,268 @@
...
+            - "loss_func": The loss function for the model training. Normally
+                speaking, this should be the same as the target function.
+            - "target_func": The target function for the data attribution. Normally
+                speaking, this should be the same as the loss function.
+            - "train_func": The training function for the model. Normally it's
+                not required if the pre-trained model checkpoints is enough
+                for the algorithm you want to benchmark.
+
+            The second tuple contains the ground truth for the benchmark,
+            the items are listed as following. It can be directly sent to the
```
tingwl0122, line 177: the items are listed where?

dattri/benchmark/load.py
```diff
@@ -0,0 +1,268 @@
...
+    models_full_count = _count_folders(
+        download_path / "benchmark" / identifier / "models_full",
+    )
+    models_full_list = [
+        download_path
+        / "benchmark"
+        / identifier
+        / "models_full"
+        / f"{i}"
+        / f"model_weights_{i}.pt"
```
tingwl0122, line 229: For our retraining script, I remember the pt's name will be `model_weights_0.pt` if we set `num_runs_per_subset` to 1. So I guess it is possible to have multiple checkpoint files in `f"{i}/` with different names.

Please correct me if I am wrong.

dattri/benchmark/load.py
```diff
@@ -0,0 +1,268 @@
...
+    models_half_count = _count_folders(
+        download_path / "benchmark" / identifier / "models_half",
+    )
+    models_half_list = [
+        download_path
+        / "benchmark"
+        / identifier
+        / "models_half"
+        / f"{i}"
+        / f"model_weights_{i}.pt"
```
tingwl0122, line 241: Same issue as above

dattri/benchmark/load.py
```diff
@@ -0,0 +1,268 @@
...
+        error_msg = f"Model {model} is not supported,\
+                      please choose from {SUPPORTED_MODELS}."
+        raise ValueError(error_msg)
+
+    if dataset not in SUPPORTED_DATASETS:
+        error_msg = f"Dataset {dataset} is not supported,\
+                      please choose from {SUPPORTED_DATASETS}."
+        raise ValueError(error_msg)
+
+    identifier = f"{dataset}_{model}"
```
xingjian-zhang, line 194: Is it possible that both `model` and `dataset` are supported but the combination is not supported?

dattri/benchmark/load.py
```diff
@@ -0,0 +1,268 @@
...
+        raise ValueError(error_msg)
+
+
+def load_benchmark(
+    model: str,
+    dataset: str,
+    metric: str,
+    download_path: str = "~/.dattri",
+    redownload: bool = False,
+) -> Tuple[Dict[str, Any], Tuple[torch.Tensor, torch.Tensor]]:
```
xingjian-zhang, line 134: nit: Given the returned object contains multiple entries, would it be more user-friendly to define a dataclass dedicated for the output? (e.g. like [`transformers.modeling_outputs.BaseModelOutput`](https://github.com/huggingface/transformers/blob/6c1d0b069de22d7ed8aa83f733c25045eea0585d/src/transformers/modeling_outputs.py#L25))

TheaperDeng: I think this should be a good idea and worth a dedicated design. I will record this for later update.

dattri/benchmark/load.py
tingwl0122: yeah, we only support limited combination of the two

TheaperDeng: will check this, thx guys

dattri/benchmark/load.py
TheaperDeng: Will support more in our huggingface repo (not in this PR)

dattri/benchmark/load.py
TheaperDeng: yes, added one and verify the performance.

dattri/benchmark/load.py
TheaperDeng: Good catch. I directly change them to `model_weights_0.pt`. These are the models trained by full/half datasets. So only 1 model will be put to each folder.

dattri/benchmark/load.py
TheaperDeng: directly remove the separate check of model and dataset. make a identifier check to handle this. Good catch, thx.

dattri/benchmark/load.py
tingwl0122: If this is not included in this PR, then LGTM now.

TheaperDeng:
merge this to keep rolling for the first release.

PR #108, opened 2024-07-18
[dattri.func] Re-define Arnoldi algorithm as a projection method

tingwl0122:
### 1. Motivation and Context
Originally, Arnoldi algorithm is defined as one of the ihvp method in `dattri/func/ihvp.py`, but funcitonal-wise, it is closer to a projection method with approximation on the inverse-Hessian. Thus, we move it to `dattri/func/projection.py`.

### 2. Summary of the change
- re-open arnoldi pr (#100)
- fix `mnist_lr` and `mnist_mlp` loss definition
- fix corresponding test files


docs/source/index.rst
```diff
@@ -29,7 +29,7 @@
 
    api/hessian.rst
    api/fisher.rst
-   api/random_projection.rst
+   api/projection.rst
```
TheaperDeng, line 32: I think you may also add
```rst
.. autofunction:: dattri.func.random_projection.arnoldi_project
```
to `projection.rst`

TheaperDeng (approved):
others LGTM

docs/source/index.rst
tingwl0122: fixed, thanks!

PR #109, opened 2024-07-19
[dattri.algorithm] Fix bugs in TRAK projection

sx-liu:
### 1. Motivation and Context

Current TRAK projection fails for large models. Since all the parameters are flattened and concatenated together, as long as the total number of parameters is larger than `max_chunk_size`, the projector will fail to arrange inputs correctly and raise an internal error.

### 2. Summary of the change

This pr changes the inputs passed to the projector from a large tensor to a dictionary and add necessary condition check for the input model.

Specifically, modified `TRAKAttributor` in `dattri.algorithm.trak`

- [x] N/A: No test will be added
This error only occurs for large models (at least > 0.1b). The error is not caused by any problems in the modules, but by the input format.

### DIFF
--- a/dattri/algorithm/trak.py
+++ b/dattri/algorithm/trak.py
@@ -63,6 +63,10 @@ class TRAKAttributor(BaseAttributor):
             projector_kwargs (Optional[Dict[str, Any]], optional): The kwargs for the
                 random projection. Defaults to None.
             device (str): The device to run the attributor. Default is cpu.
+
+        Raise:
+            ValueError: If the provided model contains layers with more parameters
+                than `max_chunk_size` of the selected cuda projector.
         """
         self.task = task
         self.norm_scaler = (
--- a/dattri/algorithm/trak.py
+++ b/dattri/algorithm/trak.py
@@ -64,7 +64,7 @@ class TRAKAttributor(BaseAttributor):
                 random projection. Defaults to None.
             device (str): The device to run the attributor. Default is cpu.
 
-        Raise:
+        Raises:
             ValueError: If the provided model contains layers with more parameters
                 than `max_chunk_size` of the selected cuda projector.
         """
--- a/dattri/algorithm/trak.py
+++ b/dattri/algorithm/trak.py
@@ -108,6 +108,21 @@ class TRAKAttributor(BaseAttributor):
                 grad_t = self.grad_func(parameters, train_batch_data)
                 grad_t = torch.nan_to_num(grad_t)
                 grad_t /= self.norm_scaler
+
+                # Unflatten the gradient tensors
+                grad_t = [
+                    _unflatten_params(g, self.task.model)
+                    for g in grad_t
+                ]
+
+                # Stack the gradient tensors for each layer
+                grad_t = {
+                    key: torch.stack(
+                        [g[key] for g in grad_t],
+                        ).view(len(grad_t), -1)
+                    for key in grad_t[0]
+                }
+
                 grad_p = (
                     random_project(
                         grad_t,
@@ -210,6 +225,20 @@ class TRAKAttributor(BaseAttributor):
                     grad_t = torch.nan_to_num(grad_t)
                     grad_t /= self.norm_scaler
 
+                    # Unflatten the gradient tensors
+                    grad_t = [
+                        _unflatten_params(g, self.task.model)
+                        for g in grad_t
+                    ]
+
+                    # Stack the gradient tensors for each layer
+                    grad_t = {
+                        key: torch.stack(
+                            [g[key] for g in grad_t],
+                            ).view(len(grad_t), -1)
+                        for key in grad_t[0]
+                    }
+
                     grad_p = (
                         random_project(
                             grad_t,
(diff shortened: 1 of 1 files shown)


tingwl0122:
Thanks for the PR, @sx-liu!
Could you also verify the performance of this TRAK version?
I assume you are attributing some larger models, does the performance (i.e., LDS) look normal for you task?

sx-liu:
> Thanks for the PR, @sx-liu! Could you also verify the performance of this TRAK version? I assume you are attributing some larger models, does the performance (i.e., LDS) look normal for you task?

I haven't tried it yet, but after running the example on mnist with some small models (lr), I found the accuracy is pretty close to the previous results.

tingwl0122:
> > Thanks for the PR, @sx-liu! Could you also verify the performance of this TRAK version? I assume you are attributing some larger models, does the performance (i.e., LDS) look normal for you task?
> 
> I haven't tried it yet, but after running the example on mnist with some small models (lr), I found the accuracy is pretty close to the previous results.
> 
> ```
> Peak memory usage: 36.384256 MB
> torch.Size([1000])
> [(0, 0), (100, 78), (200, 100), (300, 101), (400, 102), (500, 102), (600, 102), (700, 103), (800, 103), (900, 103)]
> Checked Data Sample      Found flipped Sample     
> --------------------------------------------------
> 0                        0                        
> 100                      78                       
> 200                      100                      
> 300                      101                      
> 400                      102                      
> 500                      102                      
> 600                      102                      
> 700                      103                      
> 800                      103                      
> 900                      103                      
> ```
> 
> I will try to verify it with MT5 model on a small scale of experiment.

yeah, but if this model is not large enough, then the desired functionality will not be tested right

dattri/algorithm/trak.py
```diff
@@ -108,6 +125,21 @@ def cache(
...
+                    for g in grad_t
+                ]
+
+                # Stack the gradient tensors for each layer
+                grad_t = {
+                    key: torch.stack(
+                        [g[key] for g in grad_t],
+                        ).view(len(grad_t), -1)
+                    for key in grad_t[0]
+                }
```
tingwl0122, line 141: I don't think we need this step? Eventually all the params (in each module) will be projected and added together.

dattri/algorithm/trak.py
```diff
@@ -210,6 +242,20 @@ def attribute(
...
+                        for g in grad_t
+                    ]
+
+                    # Stack the gradient tensors for each layer
+                    grad_t = {
+                        key: torch.stack(
+                            [g[key] for g in grad_t],
+                            ).view(len(grad_t), -1)
+                        for key in grad_t[0]
+                    }
```
tingwl0122, line 257: same issue as above

dattri/algorithm/trak.py
```diff
@@ -245,6 +291,20 @@ def attribute(
...
+                    for g in grad_t
+                ]
+
+                # Stack the gradient tensors for each layer
+                grad_t = {
+                    key: torch.stack(
+                        [g[key] for g in grad_t],
+                        ).view(len(grad_t), -1)
+                    for key in grad_t[0]
+                }
```
tingwl0122, line 306: same issue as above

tingwl0122 (commented):
If it is hard to get real LDS performance, we can just make sure `Chunkedcudaprojector` is called when we are attributing big models.

dattri/algorithm/trak.py
```diff
@@ -71,6 +75,19 @@ def m(params, image_label_pair):
...
+                    self.projector_kwargs["proj_max_batch_size"])
+        for p in self.task.model.parameters():
+            num_params = p.numel()
+            if num_params > max_chunk_size:
+                error_msg = (f"Your model contains layers with at least {num_params} "
+                             "parameters, which is larger than the `max_chunk_size` "
+                             f"{max_chunk_size} of the projector. This is currently "
+                             "not supported. To avoid this error, please try a smaller"
+                             " `proj_max_batch_size` or use cpu projector instead.")
+                raise ValueError(error_msg)
```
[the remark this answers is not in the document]
sx-liu: Actually even if all the inputs are passed as dict there is still an upper bound on the number of parameters in each single layer. Some layers, say the embedding layers in LLMs, can be even larger the chunk size of some projectors.

 We could remove this error after we implemented the function of further breaking the tensors down into feasible sizes.

dattri/algorithm/trak.py
sx-liu: I think it's just a convenient way to project batched gradients. Here we just stacked the gradients in each layer as `(batch_size, dim_out * dim_in)` so that the projector can process them all at once. Otherwise we may need to use a loop to process every gradient in the batch one by one.

sx-liu:
> > > Thanks for the PR, @sx-liu! Could you also verify the performance of this TRAK version? I assume you are attributing some larger models, does the performance (i.e., LDS) look normal for you task?
> > 
> > 
> > I haven't tried it yet, but after running the example on mnist with some small models (lr), I found the accuracy is pretty close to the previous results.
> > ```
> > Peak memory usage: 36.384256 MB
> > torch.Size([1000])
> > [(0, 0), (100, 78), (200, 100), (300, 101), (400, 102), (500, 102), (600, 102), (700, 103), (800, 103), (900, 103)]
> > Checked Data Sample      Found flipped Sample     
> > --------------------------------------------------
> > 0                        0                        
> > 100                      78                       
> > 200                      100                      
> > 300                      101                      
> > 400                      102                      
> > 500                      102                      
> > 600                      102                      
> > 700                      103                      
> > 800                      103                      
> > 900                      103                      
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I will try to verify it with MT5 model on a small scale of experiment.
> 
> yeah, but if this model is not large enough, then the desired functionality will not be tested right

I have tested MT5 model on ftrace dataset and obtained the MRR. The MRR is approximately 0.25 with only one checkpoint. The value is lower than the released one in the paper but I think it's reasonable because their experiment used maybe 10 checkpoints.

And I think the experiment on MNIST can also demonstrate that at least the unflattening of gradient is correct, because whichever projector we use the unflattening added in this PR is the same.

tingwl0122:
Will not follow this direction since unflattening will still cause problems if a single layer is too large (i.e. exceed the maximum chunk size).
We will have another PR to always concat the grads to tensors and split the tensors into pieces for projection if needed.

PR #111, opened 2024-07-20
[dattri.task] Add `loss_func` and make `target_func` optional

TheaperDeng:
### 1. Motivation and Context
We want to make 2 concepts clearer.

- `loss_func`: the loss function used to train the model to be attributed. Normally, this one is used to estimate the changes of parameter when one data point is up/down weighted (IF style) or between each iterations (TracIN style).
- `target_func`: the target function we want to attribute on. This one is normally set to be the same as loss function but not necessarily to be the same. so we make it optional in `dattri.task.AttributionTask`.

### 2. Summary of the change
1. add `loss_func` in `dattri.task.AttributionTask` and make `target_func` optional.
2. change some internal implementation of `dattri.algorithm` (no API change)

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,7 +34,7 @@ pip install dattri[all]
 ### Apply Data Attribution methods on PyTorch Models
 
 One can apply different data attribution methods on PyTorch Models. One only needs to define:
-1. loss function used for model training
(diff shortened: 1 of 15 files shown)


tingwl0122:
LGTM now

tingwl0122 (approved):

PR #112, opened 2024-07-20
[dattri.algorithm] Fix Arnoldi algorithm's bug

tingwl0122:
## Description

Fix Arnoldi algorithm's bug by using `arnoldi_project` defined in `dattri/projection.py`

### 1. Motivation and Context
Now the eigenspace calculated by Arnoldi algorithm will be fixed throughout the attribution.

### 2. Summary of the change
- modify`IFAttributorArnoldi` in `dattri/algorithm/influence_function.py`
- fix ruff formats in mnist examples
- modify minor designs in `dattri/func/projection.py`

### DIFF
--- a/dattri/algorithm/influence_function.py
+++ b/dattri/algorithm/influence_function.py
@@ -98,7 +98,10 @@ class IFAttributorArnoldi(BaseInnerProductAttributor):
     """The inner product attributor with Arnoldi inverse hessian transformation."""
 
     def __init__(self, *args, **kwargs) -> None:
-        """Initialize the attributor."""
+        """Initialize the attributor.
+
+        # noqa: DAR101
+        """
         super().__init__(*args, **kwargs)
         # init arnoldi projector
         self.arnoldi_projector = None
--- a/dattri/algorithm/influence_function.py
+++ b/dattri/algorithm/influence_function.py
@@ -97,15 +97,6 @@ class IFAttributorCG(BaseInnerProductAttributor):
 class IFAttributorArnoldi(BaseInnerProductAttributor):
     """The inner product attributor with Arnoldi inverse hessian transformation."""
 
-    def __init__(self, *args, **kwargs) -> None:
-        """Initialize the attributor.
-
-        # noqa: DAR101
-        """
-        super().__init__(*args, **kwargs)
-        # init arnoldi projector
-        self.arnoldi_projector = None
-
     def transformation_on_query(
         self,
         index: int,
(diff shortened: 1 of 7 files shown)


tingwl0122:
The performance is (slightly) verified through MNIST+LR example in `dattri/example/mnist_lr/influence_function_noisy_label.py`
0                        0                        
100                      54                       
200                      81                       
300                      87                       
400                      88                       
500                      88                       
600                      89                       
700                      89                       
800                      89                       
900                      89

dattri/algorithm/influence_function.py
```diff
@@ -123,14 +132,21 @@ def transformation_on_query(
...
-            partial(self.task.get_target_func(), data_target_pair=train_data),
-            **transformation_kwargs,
-        )
-        model_params, _ = self.task.get_param(index)
-        return self.ihvp_func((model_params,), query).detach()
+        from dattri.func.projection import arnoldi_project
+
+        if self.arnoldi_projector is None:
+            feature_dim = query.shape[1]
+            func = partial(self.task.get_target_func(), data_target_pair=train_data)
```
TheaperDeng, line 139: may need to use `get_loss_func` after the merge of renaming PR

TheaperDeng (commented):
others LGTM

dattri/algorithm/influence_function.py
tingwl0122: yeah, if renaming PR is merged first, I can modify this.

dattri/algorithm/influence_function.py
```diff
@@ -97,6 +97,15 @@ def transformation_on_query(
...
     """The inner product attributor with Arnoldi inverse hessian transformation."""
 
+    def __init__(self, *args, **kwargs) -> None:
+        """Initialize the attributor.
+
+        # noqa: DAR101
+        """
+        super().__init__(*args, **kwargs)
+        # init arnoldi projector
+        self.arnoldi_projector = None
```
tingwl0122, line 107: actually, I am wondering whether it is possible to make this more clean. @TheaperDeng do you have any thought?

TheaperDeng: I guess this is just for the code style checking right?

tingwl0122: yeah, but I mean whether it is possible to append one more variable (the projector here) without writing all these stuffs

TheaperDeng: I think for users who are using `dattri`, they may directly forget all these thing since they won't need the code style checking.

For us, I guess some options are
1. just ignore this code style checking rule
2. Use property https://www.programiz.com/python-programming/property

tingwl0122: Use other workaround to avoid this `init` overriding

dattri/algorithm/influence_function.py
tingwl0122: fixed.

TheaperDeng (approved):

PR #113, opened 2024-07-21
[dattri.func] Fix `ChunkedCudaProjector` bug

tingwl0122:
### 1. Motivation and Context
We found that `ChunkedCudaProjector` not working as expected for large models. For Dict input, we found it hard to project without dividing gradients, so we will inherently force dict input to be tensor and split the tensor for projection if it is too large.

### 2. Summary of the change
- fix chunkcudaprojector bug by forcing tensor input
- add chunkcudaprojector test case for tensor input

Note that I preserve most of the original code and possibly we can come back and make it support dict input (without enforcing it to be tensor).

### DIFF
--- a/dattri/func/projection.py
+++ b/dattri/func/projection.py
@@ -573,7 +573,7 @@ class ChunkedCudaProjector:
                 ),
             )
 
-        return ch_output[: self.feature_batch_size]
+        return ch_output
 
 
 class ArnoldiProjector(AbstractProjector):
--- a/dattri/func/projection.py
+++ b/dattri/func/projection.py
@@ -457,7 +457,7 @@ class ChunkedCudaProjector:
         del self.ch_input
         self.input_allocated = False
 
-    def project(self, features: Union[dict, Tensor], ensemble_id: int) -> Tensor:
+    def dict_project(self, features: Union[dict, Tensor], ensemble_id: int) -> Tensor:
         """Performs the random projection on the feature matrix.
 
         Args:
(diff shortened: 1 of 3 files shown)


dattri/func/projection.py
```diff
@@ -457,7 +457,7 @@ def free_memory(self) -> None:
         del self.ch_input
         self.input_allocated = False
 
-    def project(self, features: Union[dict, Tensor], ensemble_id: int) -> Tensor:
+    def dict_project(self, features: Union[dict, Tensor], ensemble_id: int) -> Tensor:
```
TheaperDeng, line 460: will this `dict_project` be used later?

dattri/func/utils.py
```diff
@@ -88,7 +88,7 @@ def _vectorize(
     return arr
 
 
-def get_parameter_chunk_sizes(
+def _get_parameter_chunk_sizes(
```
TheaperDeng, line 91: will this function be used later?

dattri/func/projection.py
tingwl0122: yeah, probably. If we are going to support dict input

dattri/func/utils.py
tingwl0122: as above, might be used together

TheaperDeng (approved):
LGTM, please also have a look @sx-liu

sx-liu (approved):
Also LGTM, works for MT5 model.

PR #116, opened 2024-07-23
[Doc] Quick fix on the document about `__init__` methods' correct presentation.

TheaperDeng:
## Description

This a quick fix to make `__init__` shown in our document.


TheaperDeng:
This is a rather small pr with only document fixing. I will directly merge it.

PR #118, opened 2024-07-28
[dattri.task] add partial parameter support for target/loss function and their gradient func

TheaperDeng:
### 1. Motivation and Context
1. This PR supports some extensive requirements for calculating the loss/target function and their grad/hessian based on only a portion of the parameter (e.g., some specific layers).
2. The support should be completely wrapped inside the task abstraction.

### 2. Summary of the change
1. Support the `layer_name` for `task.get_grad_target_func`, `task.get_target_func`, `task.get_grad_loss_func`, and `task.get_loss_func` (previously it will raise a notimplemented error).

TODO: this is the first PR to support partial parameter in dattri. Next PRs will support this feature in high-level attributor APIs.

```python
task = AttributionTask(loss_func=f, model=model, checkpoints=model.state_dict())

# only use the last layer's parameter
grad_func_partial = task.get_grad_loss_func(
    layer_name=["fc3.weight", "fc3.bias"],
    index=0,  # this is a special additional parameter, because some other parameters we need to copy from the checkpoint.
)
params_partial, _ = task.get_param(
    layer_name=["fc3.weight", "fc3.bias"],
    index=0,
)
gradient_partial = grad_func_partial(params_partial, data_pair)

# only use the full parameter
grad_func_full = task.get_grad_loss_func()
params_full, _ = task.get_param()
gradient_full = grad_func_full(params_full, data_pair)

# the last layer's parameter's gradient should be the same as the last few dim full parameter's gradient.
torch.all_close(gradient_full[-last_layer_index:], gradient_partial)

```

### DIFF
--- a/dattri/task.py
+++ b/dattri/task.py
@@ -14,7 +14,7 @@ import torch
 from torch import nn
 from torch.func import grad, vmap
 
-from dattri.func.utils import flatten_func, flatten_params
+from dattri.func.utils import flatten_func, flatten_params, partial_param
 
 
 class AttributionTask:
@@ -97,9 +97,17 @@ class AttributionTask:
 
         # TODO: Make this more general, that is allow customized kwargs.
         self.grad_loss_func = vmap(grad(self.loss_func), in_dims=(None, 1))
-        self.grad_loss_func_kwargs = {"in_dims": (None, 1)}
+        self.grad_loss_func_kwargs = {
+            "in_dims": (None, 1),
+            "layer_name": None,
+            "index": None,
+        }
         self.grad_target_func = vmap(grad(self.target_func), in_dims=(None, 1))
-        self.grad_target_func_kwargs = {"in_dims": (None, 1)}
+        self.grad_target_func_kwargs = {
+            "in_dims": (None, 1),
+            "layer_name": None,
+            "index": None,
+        }
 
     def _load_checkpoints(self, index: int) -> None:
         """This method load the checkpoint at the specified index.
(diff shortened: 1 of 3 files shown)


sx-liu (approved):
LGTM. I tried with tracin by modifying some of the source codes of TracInAttributor.

charles-pyj:
LGTM as well. I tried the gradients on both linear and convolution layers.

charles-pyj:
But do we want to also add merging for the parameters for the same layer as well? For example, if one wants to conduct attribution on the first linear layer, it would be handy if we can combine the linear1.weight and linear1.bias together and obtain the concatenated gradients other than getting the separate gradients first and concatenate them later in attribution calculation?

charles-pyj (approved):

TheaperDeng:
> But do we want to also add merging for the parameters for the same layer as well? For example, if one wants to conduct attribution on the first linear layer, it would be handy if we can combine the linear1.weight and linear1.bias together and obtain the concatenated gradients other than getting the separate gradients first and concatenate them later in attribution calculation?

I thought about this as well. Since we are using the `named_parameters().keys()` as the valid choices for layer names. I do afraid some times the layer's parameter is not named as `xxx.weight` and `xxx.bias`, then it may bring some error.

jiaqima (approved):
LGTM

dattri/task.py
```diff
@@ -168,26 +177,36 @@ def get_grad_target_func(
...
 
         Returns:
             Callable: The function that computes the gradient of the target function.
-
-        Raises:
-            NotImplementedError: If layer_name is not None.
         """
+        # first add decorator that handles the layer_name
+        target_func = self.target_func
         if layer_name is not None:
```
jiaqima, line 188: Here the condition is "not None" while in the function `get_param`, the default value of `layer_name` is 0.  We need to make it consistent.

jiaqima: Also, do we want to support a list of index?

PR #121, opened 2024-08-09
[Document] update the Metrics Supported

SeanZh30:
Add brittleness test for Metrics Supported

## Description

Add brittleness test for mertics supported

### 1. Motivation and Context

Update description for let user know the new metrics

### 2. Summary of the change

Update readme.md file

NA


TheaperDeng:
Feel free to make this change in the brittleness PR rather than in a separate one.

PR #122, opened 2024-08-12
[dattri.algorithm] Add KNN data shapley

sx-liu:
### 1. Motivation and Context

To include the data shapley methods as attribution methods as a completion of the attribution algorithms. Data shapley methods are game-theoretic methods that assign the "relative value of data". This PR focuses on the data shapley values related to KNN.

### 2. Summary of the change

1. Add new file `dattri.algorithm.data_shapley` and `KNNShalpeyAttributerExact` as the first data shapley value attributer.
2. Add test case for `KNNShalpeyAttributerExact`.

### DIFF
--- a/dattri/algorithm/data_shapley.py
+++ b/dattri/algorithm/data_shapley.py
@@ -40,7 +40,7 @@ def default_dist_func(
 
 
 class KNNShalpeyAttributor(BaseAttributor):
-    """KNN Data Shapley Attributer."""
+    """KNN Data Shapley Attributor."""
 
     def __init__(
         self,
--- a/dattri/algorithm/data_shapley.py
+++ b/dattri/algorithm/data_shapley.py
@@ -6,7 +6,7 @@ from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
     from collections.abc import Callable
-    from typing import List
+    from typing import List, Optional, Tuple
 
 import torch
 from tqdm import tqdm
@@ -15,9 +15,22 @@ from .base import BaseAttributor
 from .utils import _check_shuffle
 
 
-def default_dist_func(train_batch, test_batch):
-    coord1 = train_batch[0]
-    coord2 = test_batch[0]
+def default_dist_func(
+        batch_x: Tuple[torch.Tensor],
+        batch_y: Tuple[torch.Tensor],
+) -> torch.Tensor:
+    """Default distance function for KNN.
+
+    Args:
+        batch_x (Tuple[torch.Tensor]): The batch to be calculated
+            distances on. The embeddings is default to be the
+            first element.
+        batch_y (Tuple[torch.Tensor]): The batch to be calculated
+            distances on. The embeddings is default to be the
+            first element.
+    """
+    coord1 = batch_x[0]
+    coord2 = batch_y[0]
     return torch.cdist(coord1, coord2)
 
 
@@ -27,7 +40,7 @@ class KNNShalpeyAttributerExact(BaseAttributor):
     def __init__(
         self,
         k_neighbors: int,
-        distance_func: Callable = None,
+        distance_func: Optional[Callable] = None,
     ) -> None:
         """Initialize the AttributionTask.
 
@@ -37,17 +50,19 @@ class KNNShalpeyAttributerExact(BaseAttributor):
 
         Args:
             k_neighbors (int): The number of neighbors in KNN model.
-            distance_func (Callable): Customizable function used for
-                distance calculation in KNN. The function can be quite
-                flexible in terms of what is calculated, but it should
-                take two batches of data as input.
+            distance_func (Callable, optional): Customizable function
+                used for distance calculation in KNN. The function
+                can be quite flexible in terms of what is calculated,
+                but it should take two batches of data as input.
                 A typical example is as follows:
                 ```python
-                def f(train_batch, test_batch):
-                    coord1 = train_batch[0]
-                    coord2 = test_batch[0]
+                def f(batch_x, batch_y):
+                    coord1 = batch_x[0]
+                    coord2 = batch_y[0]
                     return torch.cdist(coord1, coord2)
                 ```.
+                If not provided, a default Euclidean distance function
+                will be used.
         """
         self.k_neighbors = k_neighbors
 
--- a/dattri/algorithm/data_shapley.py
+++ b/dattri/algorithm/data_shapley.py
@@ -72,6 +72,9 @@ class KNNShalpeyAttributor(BaseAttributor):
                 ```.
                 If not provided, a default Euclidean distance function
                 will be used.
+
+        Raises:
+            NotImplementedError: If task is not None.
         """
         self.k_neighbors = k_neighbors
 
--- a/dattri/algorithm/data_shapley.py
+++ b/dattri/algorithm/data_shapley.py
@@ -37,7 +37,7 @@ def default_dist_func(
     return torch.cdist(coord1, coord2)
 
 
-class KNNShalpeyAttributerExact(BaseAttributor):
+class KNNShalpeyAttributor(BaseAttributor):
     """KNN Data Shapley Attributer."""
 
     def __init__(
--- a/dattri/algorithm/data_shapley.py
+++ b/dattri/algorithm/data_shapley.py
@@ -1,0 +1,135 @@
+"""This module implement data shapley valuation."""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from collections.abc import Callable
+    from typing import List
+
+import torch
+from tqdm import tqdm
+
+from .base import BaseAttributor
+from .utils import _check_shuffle
+
+
+class KNNShalpeyAttributerExact(BaseAttributor):
+    """KNN Data Shapley Attributer."""
+
+    def __init__(
+        self,
+        k: int,
+        distance_func: Callable,
+    ) -> None:
+        """Initialize the AttributionTask.
+
+        KNN Data Shapley Valuation is task-agnostic.
+
+        Args:
+            k (int): The number of neighbors in KNN model.
+            distance_func (Callable): Customizable function used for
+                distance calculation in KNN. The function can be quite
+                flexible in terms of what is calculated, but it should
+                take two batches of data as input.
+                A typical example is as follows:
+                ```python
+                def f(train_batch, test_batch):
+                    coord1 = train_batch[0]
+                    coord2 = test_batch[0]
+                    return torch.cdist(coord1, coord2)
+                ```.
+        """
+        self.k = k
+        self.distance_func = distance_func
+
+    def cache(self) -> None:
+        """Precompute and cache some values for efficiency."""
+
+    def attribute(
+        self,
+        train_dataloader: torch.utils.data.DataLoader,
+        test_dataloader: torch.utils.data.DataLoader,
+        train_labels: List[int],
+        test_labels: List[int],
+    ) -> None:
+        """Calculate the shapley values of the training set on each test sample.
+
+        Args:
+            train_dataloader (torch.utils.data.DataLoader): The dataloader for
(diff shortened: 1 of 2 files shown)


dattri/algorithm/data_shapley.py
```diff
@@ -0,0 +1,135 @@
...
+    """KNN Data Shapley Attributer."""
+
+    def __init__(
+        self,
+        k: int,
+        distance_func: Callable,
+    ) -> None:
+        """Initialize the AttributionTask.
+
+        KNN Data Shapley Valuation is task-agnostic.
```
jiaqima, line 28: Task agnostic or model agnostic? I think the attribution/shapley values are still different for different test data points?

jiaqima (approved):
LGTM

dattri/algorithm/data_shapley.py
```diff
@@ -0,0 +1,135 @@
...
+                A typical example is as follows:
+                ```python
+                def f(train_batch, test_batch):
+                    coord1 = train_batch[0]
+                    coord2 = test_batch[0]
+                    return torch.cdist(coord1, coord2)
+                ```.
+        """
+        self.k = k
+        self.distance_func = distance_func
```
jiaqima, line 45: One more suggestion is that we implement a default Euclidean distance function when distance_func=None. This is likely ok for a lot of use cases and reduce some user burden.

dattri/algorithm/data_shapley.py
```diff
@@ -0,0 +1,135 @@
...
+
+import torch
+from tqdm import tqdm
+
+from .base import BaseAttributor
+from .utils import _check_shuffle
+
+
+class KNNShalpeyAttributerExact(BaseAttributor):
+    """KNN Data Shapley Attributer."""
```
tingwl0122, line 19: Attributor

dattri/algorithm/data_shapley.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def __init__(
+        self,
+        k: int,
```
tingwl0122, line 23: possibly call it "k_neighbors" or something similar? just to make this variable a bit more informative

dattri/algorithm/data_shapley.py
```diff
@@ -0,0 +1,135 @@
...
+                take two batches of data as input.
+                A typical example is as follows:
+                ```python
+                def f(train_batch, test_batch):
+                    coord1 = train_batch[0]
+                    coord2 = test_batch[0]
+                    return torch.cdist(coord1, coord2)
+                ```.
+        """
+        self.k = k
```
tingwl0122, line 44: same issue as above

dattri/algorithm/data_shapley.py
jiaqima: In fact, one way to make KNN Shapley to be model dependent is to use the model hidden representation to calculate the distance between two data points.

Maybe we could add a task argument in __init__ so that we could extract some hidden representations of the model for distance calculation?

dattri/algorithm/data_shapley.py
```diff
@@ -0,0 +1,135 @@
...
+
+    def cache(self) -> None:
+        """Precompute and cache some values for efficiency."""
+
+    def attribute(
+        self,
+        train_dataloader: torch.utils.data.DataLoader,
+        test_dataloader: torch.utils.data.DataLoader,
+        train_labels: List[int],
+        test_labels: List[int],
```
tingwl0122, line 55: labels should already be included in the dataloaders? why need additional input here

dattri/algorithm/data_shapley.py
```diff
@@ -0,0 +1,135 @@
...
+    ) -> None:
+        """Calculate the shapley values of the training set on each test sample.
```
tingwl0122, line 57: "kNN shapley value" to be more specific

dattri/algorithm/data_shapley.py
```diff
@@ -0,0 +1,135 @@
...
+            test_dataloader (torch.utils.data.DataLoader): The dataloader for
+                test samples to calculate the shapley values. The dataloader
+                should not be shuffled.
+            train_labels: (List[int]): The list of training labels, with the same
+                size and order of the training dataset.
+            test_labels: (List[int]): The list of test labels, with the same
+                size and order of the test dataset.
+
+        Returns:
+            Tensor: The shapley values of the training set on the test set, with
```
tingwl0122, line 72: again, the kNN shapley value

dattri/algorithm/data_shapley.py
```diff
@@ -0,0 +1,135 @@
...
+                    len(train_dataloader.sampler),
+                )
+
+                row_st = test_batch_idx * test_dataloader.batch_size
+                row_ed = min(
+                    (test_batch_idx + 1) * test_dataloader.batch_size,
+                    len(test_dataloader.sampler),
+                )
+
+                dist_matrix[row_st:row_ed, col_st:col_ed] += partial_dist
```
tingwl0122, line 114: do we need to take batch size into consideration when computing this distance matrix?
it should be faster to directly call `self.distance_func` on the two lists?
(although this part should be quite fast even computed in this batch-style)

dattri/algorithm/data_shapley.py
sx-liu: I have currently updated the distance function with an Euclidean default. I would change the default to embedding distance after the `register_forward_hook` function is implemented.

tingwl0122:
Is it possible to have a quick test on small-scale experiments to validate the correctness of the algorithm?
Something like noisy label detection for MNIST+LR on (train num, test num) = (5000, 500) should be good

sx-liu:
> Is it possible to have a quick test on small-scale experiments to validate the correctness of the algorithm? Something like noisy label detection for MNIST+LR on train num=5000 should be good

I experimented on MNIST mislabel detection with num=1000 and num=5000 separately. I used the pixel distance directly. The `k_neighbors` is chosen to be 3.

```
0                        0
100                      73
200                      97
300                      99
400                      100
500                      100
600                      100
700                      100
800                      100
900                      100
```

```
0                        0
100                      97
200                      195
300                      281
400                      368
500                      432
600                      473
700                      484
800                      492
900                      496
1000                     496
1100                     497
1200                     498
1300                     499
1400                     499
1500                     499
1600                     500
1700                     500
1800                     500
1900                     500
2000                     500
2100                     500
2200                     500
2300                     500
2400                     500
2500                     500
2600                     500
2700                     500
2800                     500
2900                     500
3000                     500
3100                     500
3200                     500
3300                     500
3400                     500
3500                     500
3600                     500
3700                     500
3800                     500
3900                     500
4000                     500
4100                     500
4200                     500
4300                     500
4400                     500
4500                     500
4600                     500
4700                     500
4800                     500
4900                     500
```

dattri/algorithm/data_shapley.py
sx-liu: Actually I am concerned about the memory issue of calculating the distance matrix, because the data can be something other than tensor, and the user may also include some preprocessing in their distance computation. I don't know whether concatenating everything into a list would cause oom issues.

dattri/algorithm/data_shapley.py
jiaqima: Sounds good. We should also change the description in the docstring here accordingly.

dattri/algorithm/data_shapley.py
```diff
@@ -0,0 +1,178 @@
...
+            distances on. The embeddings is default to be the
+            first element.
+        batch_y (Tuple[torch.Tensor]): The batch to be calculated
+            distances on. The embeddings is default to be the
+            first element.
+
+    Returns:
+        Tensor: The calculated Euclidean distance.
+    """
+    coord1 = batch_x[0]
```
jiaqima, line 37: We are assuming the dataloader will yield batches in the form of (x, y) pair here.

In this case, maybe we could remove train_labels and test_labels in the self.attribute() function, and add a couple of value errors if the dataloader does not follow this format.

dattri/algorithm/data_shapley.py
jiaqima: The batch version sounds good to me.

tingwl0122: sure, then it should better to just keep the batch version.

tingwl0122:
> > Is it possible to have a quick test on small-scale experiments to validate the correctness of the algorithm? Something like noisy label detection for MNIST+LR on train num=5000 should be good
> 
> I experimented on MNIST mislabel detection with num=1000 and num=5000 separately. I used the pixel distance directly. The `k_neighbors` is chosen to be 3.
> 
> ```
> Checked Data Sample      Found flipped Sample
> --------------------------------------------------
> 0                        0
> 100                      73
> 200                      97
> 300                      99
> 400                      100
> 500                      100
> 600                      100
> 700                      100
> 800                      100
> 900                      100
> ```
> 
> ```
> Checked Data Sample      Found flipped Sample
> --------------------------------------------------
> 0                        0
> 100                      97
> 200                      195
> 300                      281
> 400                      368
> 500                      432
> 600                      473
> 700                      484
> 800                      492
> 900                      496
> 1000                     496
> 1100                     497
> 1200                     498
> 1300                     499
> 1400                     499
> 1500                     499
> 1600                     500
> 1700                     500
> 1800                     500
> 1900                     500
> 2000                     500
> 2100                     500
> 2200                     500
> 2300                     500
> 2400                     500
> 2500                     500
> 2600                     500
> 2700                     500
> 2800                     500
> 2900                     500
> 3000                     500
> 3100                     500
> 3200                     500
> 3300                     500
> 3400                     500
> 3500                     500
> 3600                     500
> 3700                     500
> 3800                     500
> 3900                     500
> 4000                     500
> 4100                     500
> 4200                     500
> 4300                     500
> 4400                     500
> 4500                     500
> 4600                     500
> 4700                     500
> 4800                     500
> 4900                     500
> ```

Thanks! Looks great!

dattri/algorithm/data_shapley.py
tingwl0122: If we follow this, then one of my above concern can also be resolved.

sx-liu: I have updated by making the labels optional.

dattri/algorithm/data_shapley.py
```diff
@@ -0,0 +1,205 @@
...
+
+                row_st = test_batch_idx * test_dataloader.batch_size
+                row_ed = min(
+                    (test_batch_idx + 1) * test_dataloader.batch_size,
+                    len(test_dataloader.sampler),
+                )
+
+                dist_matrix[row_st:row_ed, col_st:col_ed] += partial_dist
+
+                if len(train_labels) != len(train_dataloader.sampler):
```
jiaqima, line 180: Add a comment here that we are assuming train_batch_data is a tuple with the last element as the batch of train_labels

sx-liu: Done.

dattri/algorithm/data_shapley.py
```diff
@@ -0,0 +1,207 @@
...
+    return torch.cdist(coord1, coord2)
+
+
+class KNNShalpeyAttributor(BaseAttributor):
+    """KNN Data Shapley Attributer."""
+
+    def __init__(
+        self,
+        k_neighbors: int,
+        task: AttributionTask = None,
```
TheaperDeng, line 48: Please add an issue to the issue page, where we can track the potential improvement. One design I am think is that maybe we can left `distance_func` visit the `task` and extract what ever they want.

```python
def distance_func(batch_x, batch_y, task=None):
    ...
```

TheaperDeng (approved):
The PR LGTM. Please revise some typo @tingwl0122 mentioned earlier

dattri/algorithm/data_shapley.py
sx-liu: Please check issue #124

dattri/algorithm/data_shapley.py
jiaqima: Please address this minor comment for the docstring.

PR #123, opened 2024-08-15
[dattri.algorithm] Changes to the implementation of DataInf.

charles-pyj:
## Description

Changes to the implementation of DataInf

### 1. Motivation and Context

Previous implementation is based on ifvps, which causes heavy memory overhead together with the use of vmap because in total (num_train,parameter,num_val) tensor is stored in memory. 

### 2. Summary of the change

- Remove the dependency of datainf on ifvp, instead directly calculate the influence values for smaller tensors saved in memory.
- Deleted old ifvp datainf implementations
- Added `IFAttributorDataInf ` to `dattri.algorithm.influence_function.py`
- MLP+MNIST: before: Memory usage: 26059M, Time usage: 1206s, now: 6821M, Time usage: 979s
- [x] Testing transform_test_rep


dattri/algorithm/influence_function.py
```diff
@@ -5,15 +5,23 @@
...
     from torch import Tensor
+    from torch.utils.data import DataLoader
 
+    from dattri.task import AttributionTask
+
+import warnings
 from functools import partial
 
 import torch
+from torch.func import grad
```
jiaqima, line 20: How about just use torch.func.grad when needed? Since vmap is used in this way.

dattri/algorithm/influence_function.py
```diff
@@ -201,59 +209,416 @@ def lissa_collate_fn(
         )
 
 
-class IFAttributorDataInf(BaseInnerProductAttributor):
-    """The inner product attributor with DataInf inverse hessian transformation."""
+class IFAttributorDataInf(BaseAttributor):
```
jiaqima, line 212: Maybe it's better to still inherit from BaseInnerProductAttributor and just overwrite some of the member functions, given a lot of member functions are the same here.

dattri/algorithm/influence_function.py
```diff
@@ -201,59 +209,416 @@ def lissa_collate_fn(
...
-            (None, 0),
-            param_layer_map=param_layer_map,
-            **transformation_kwargs,
+        Args:
+            full_train_dataloader (DataLoader): The dataloader
+                with full training samples for inverse hessian calculation.
+        """
+        self._set_full_train_data(full_train_dataloader)
+
+    @staticmethod
```
jiaqima, line 355: Why staticmethod?

jiaqima (changes requested):

dattri/algorithm/influence_function.py
charles-pyj: Fixed.

dattri/algorithm/influence_function.py
charles-pyj: Because the ihvp method does not alter anything of the class itself and at the same time I want to make it a class function.

dattri/algorithm/influence_function.py
charles-pyj: Fixed.

TheaperDeng (changes requested):
I give a review to the changes and it should be correct. The only concern is that this breaks the original purpose to create a `BaseInnerProductAttributor`.

The original purpose is that users could only implement some methods like `transformation_on_query`. But this implementation may reveal some problems may lie in the current implementation `attribute`. 

**Please fix the document and the conflict so that we can merge this PR.**

charles-pyj:
> I give a review to the changes and it should be correct. The only concern is that this breaks the original purpose to create a `BaseInnerProductAttributor`.
> 
> The original purpose is that users could only implement some methods like `transformation_on_query`. But this implementation may reveal some problems may lie in the current implementation `attribute`. (I guess we can try to fix the base class later)
> 
> **Please fix the document and the conflict so that we can merge this PR.**

@TheaperDeng Thanks! Sorry for the late update. After discussion we have found another way that solves the memory issue but does not require implementing Datainf separately(as this PR does). I will keep that in another PR(or reopen this PR later when I finish) and for now let's just use this PR as a stand-alone implementation for testing Datainf? Plus the conflicts and ruff checks are solved.

dattri/algorithm/influence_function.py
```diff
@@ -405,14 +406,14 @@ def lissa_collate_fn(
...
-    """The inner product attributor with DataInf inverse hessian transformation."""
+    """The attributor using DataInf."""
 
     def transformation_on_query(
-        self,
-        index: int,
-        train_data: Tuple[torch.Tensor, ...],
-        query: torch.Tensor,
-        **transformation_kwargs,
+    self,
```
jiaqima, line 412: Are these changes of indent intended?

charles-pyj: Oh they are unintended. Fixed

dattri/algorithm/influence_function.py
```diff
@@ -433,30 +433,200 @@ def transformation_on_query(
             torch.Tensor: The transformation on the query. Normally it is a 2-d
                 dimensional tensor with the shape of (batchsize, transformed_dimension).
         """
-        from dattri.func.fisher import ifvp_datainf
 
-        model_params, param_layer_map = self.task.get_param(index, layer_split=True)
+        def _single_datainf(
```
jiaqima, line 437: maybe name it as _transform_single_test_rep?

dattri/algorithm/influence_function.py
```diff
@@ -433,30 +433,200 @@ def transformation_on_query(
...
+            v: torch.Tensor,
+            grad: torch.Tensor,
+            regularization: float,
+        ) -> torch.Tensor:
+            """Intermediate DataInf value calculation of a single data point.
+
+            Args:
+                v (torch.Tensor): A tensor representing (batched) validation
+                    set gradient,Normally of shape (num_valiation,parameter_size)
```
jiaqima, line 446: typo

dattri/algorithm/influence_function.py
```diff
@@ -433,30 +433,200 @@ def transformation_on_query(
...
         """
-        from dattri.func.fisher import ifvp_datainf
 
-        model_params, param_layer_map = self.task.get_param(index, layer_split=True)
+        def _single_datainf(
+            v: torch.Tensor,
+            grad: torch.Tensor,
+            regularization: float,
+        ) -> torch.Tensor:
+            """Intermediate DataInf value calculation of a single data point.
```
jiaqima, line 442: More detailed description. Better include the markdown format of the formula of this function

dattri/algorithm/influence_function.py
```diff
@@ -433,30 +433,200 @@ def transformation_on_query(
...
+
+            Args:
+                v (torch.Tensor): A tensor representing (batched) validation
+                    set gradient,Normally of shape (num_valiation,parameter_size)
+                grad (torch.Tensor): A tensor representing a single training
+                    gradient, of shape (parameter_size,)
+                regularization (float): A float or list of floats default
+                    to 0.0.Specifies the regularization term to be added to
```
jiaqima, line 450: typo

dattri/algorithm/influence_function.py
```diff
@@ -433,30 +433,200 @@ def transformation_on_query(
...
+                    the Hessian matrix in each layer.
```
jiaqima, line 451: Hessian matrix -> empirical Fisher information matrix

dattri/algorithm/influence_function.py
jiaqima: Please go through all the docstrings to correct typos.

jiaqima (commented):
I didn't go through the full code. It seems that the current implementation will have some serious compatibility issue with the new API design outlined in #130. So we'd better wait for that to be finalized. 

Meanwhile, there are quite a few typos in the docstrings. Please pay attention to them in the next iteration.

charles-pyj:
> I didn't go through the full code. It seems that the current implementation will have some serious compatibility issue with the new API design outlined in #130. So we'd better wait for that to be finalized.
> 
> Meanwhile, there are quite a few typos in the docstrings. Please pay attention to them in the next iteration.

Thanks. I will refine the code when the new API design is settled.

charles-pyj:
Resolved conflicts with upstream and updated docstring.

dattri/algorithm/influence_function.py
```diff
@@ -606,49 +606,231 @@ def __init__(
...
+        return query_split
 
-        vector_product = 0
-        for full_data_ in self.full_train_dataloader:
-            # move to device
-            full_data = tuple(data.to(self.device) for data in full_data_)
-            res = self.ihvp_func(
-                (model_params, (full_data[0], full_data[1].view(-1, 1).float())),
-                query_split,
+    def attribute(
```
jiaqima, line 717: Why can't we just change the functions for generating and transforming representations and reuse the original attribute function?

dattri/func/fisher.py
```diff
@@ -201,7 +201,21 @@ def _single_datainf_ifvp(
         grad: torch.Tensor,
         regularization: float,
     ) -> torch.Tensor:
-        # TODO: docstring
+        """Intermediate DataInf ifvp calculation of a single training data point.
```
jiaqima, line 204: I think we could get rid of ifvp_datainf now?

charles-pyj: Do you mean completely delete the implementations of ifvp_datainf and at_x versions

dattri/algorithm/influence_function.py
charles-pyj: This is because I need to calculate full training data gradient before I start calculating transformed queries. Also this should be done only once for all batches of training and testing data. If I were to put this in one of the transformations the gradients of full_training_data is taken multiple times and leads to slow performance.

dattri/func/fisher.py
charles-pyj: Got it!

dattri/algorithm/influence_function.py
jiaqima: It should be done in .cache then. See the Arnoldi attributor for example.

dattri/algorithm/influence_function.py
```diff
@@ -606,49 +606,231 @@ def __init__(
     def transform_test_rep(
         self,
         ckpt_idx: int,
+        full_train_rep: torch.Tensor,
```
jiaqima, line 609: You could cache full_train_rep as something like self.full_train_rep when calling self.cache()

charles-pyj: Got it.

dattri/algorithm/influence_function.py
charles-pyj: Thanks. Will follow Arnoldi for caching.

dattri/algorithm/influence_function.py
charles-pyj: @jiaqima One issue is the caching of full_train_rep requires the ckpt_idx, and adding ckpt_idx would affect how attributor is used. I can make ckpt_idx to be optional though and when user does not give that parameter only loader is set and inside the attribute loop for each ckpt full_train_rep is cached once. Does that sounds reasonable?

jiaqima: The total number of ckpt_idx could be determined by self.task. You could then iterate through all the ckpt_idx in self.cache() and cache them in a dictionary.

jiaqima: In addition, for very large models, I think we should provide an option that estimate FIM with a subset of training data. In this case, you may want to pass an argument like the `precompute_data_ratio` in Arnoldi attributor in the __init__ of this attributor. During cache, you'll only cache subsets.

You may call it self._cached_train_reps instead of full_train_rep to make the name more accurate.

charles-pyj: Got it! Actually there are some potential issues in the implementation of precompute ratio in Arnoldi and I will raise that in an issue later.

dattri/func/fisher.py
charles-pyj: The current unit test involves test of ifvps as well and I can remove them after the test is updated?

dattri/algorithm/influence_function.py
charles-pyj: Oh in that case I can use arnoldi's sample calculation instead of what I am using now. Going to fix it soon

dattri/algorithm/influence_function.py
```diff
@@ -620,35 +649,94 @@ def transform_test_rep(
...
+            Returns:
+                torch.Tensor: A tensor corresponding to transformed test representation.
+            """
+            grad = grad.unsqueeze(-1)
+            coef = (v @ grad) / (regularization + torch.norm(grad) ** 2)
+            return (v - coef @ grad.T) / regularization
+
+        regularization = self.transformation_kwargs["regularization"]
+        # Split layer-wise train and test representations
+        query_split = self.get_layer_wise_reps(ckpt_idx, test_rep)
```
jiaqima, line 686: change "query" to "test_rep"?

jiaqima: Also, split is a bit ambiguous as it usually refers to data split. How about test_rep_layers?

dattri/algorithm/influence_function.py
```diff
@@ -620,35 +649,94 @@ def transform_test_rep(
...
         """
-        from dattri.func.fisher import ifvp_datainf
 
-        model_params, param_layer_map = self.task.get_param(ckpt_idx, layer_split=True)
+        def _transform_single_test_rep(
+            v: torch.Tensor,
+            grad: torch.Tensor,
+            regularization: float,
+        ) -> torch.Tensor:
+            """Transformation of a single test representation.
```
jiaqima, line 658: ... for a single layer

dattri/algorithm/influence_function.py
```diff
@@ -620,35 +649,94 @@ def transform_test_rep(
...
+                    reg,
+                ),
+                )(batch)
+                # Sum up the contributions by chunk and average at the end
+                running_contributions += contribution.sum(dim=0)
+            running_contributions /= length
+            ifvps.append(running_contributions)
+        return torch.cat(ifvps, dim=1)
+
+    def get_layer_wise_reps(self,
```
jiaqima, line 716: How about name this as _get_layer_wise_reps since this is a minor helper function

dattri/algorithm/influence_function.py
```diff
@@ -620,35 +649,94 @@ def transform_test_rep(
...
+            """
+            grad = grad.unsqueeze(-1)
+            coef = (v @ grad) / (regularization + torch.norm(grad) ** 2)
+            return (v - coef @ grad.T) / regularization
+
+        regularization = self.transformation_kwargs["regularization"]
+        # Split layer-wise train and test representations
+        query_split = self.get_layer_wise_reps(ckpt_idx, test_rep)
+        full_rep_split = self.get_layer_wise_reps(ckpt_idx,
+         self._cached_train_reps[f"ckpt_{ckpt_idx}"])
```
jiaqima, line 688: more indent

dattri/algorithm/influence_function.py
```diff
@@ -620,35 +649,94 @@ def transform_test_rep(
...
+                torch.Tensor: A tensor corresponding to transformed test representation.
+            """
+            grad = grad.unsqueeze(-1)
+            coef = (v @ grad) / (regularization + torch.norm(grad) ** 2)
+            return (v - coef @ grad.T) / regularization
+
+        regularization = self.transformation_kwargs["regularization"]
+        # Split layer-wise train and test representations
+        query_split = self.get_layer_wise_reps(ckpt_idx, test_rep)
+        full_rep_split = self.get_layer_wise_reps(ckpt_idx,
```
jiaqima, line 687: cached_train_rep_layers

dattri/algorithm/influence_function.py
```diff
@@ -620,35 +649,94 @@ def transform_test_rep(
...
+         self._cached_train_reps[f"ckpt_{ckpt_idx}"])
+        layer_cnt = len(full_rep_split)
+        ifvps = []
```
jiaqima, line 690: transformed_test_rep_layers?

dattri/algorithm/influence_function.py
```diff
@@ -620,35 +649,94 @@ def transform_test_rep(
...
+        # Use test batch size as intermediate batch size
+        # Peak memory usage: max(train_batch_size,test_batch_size)*test_batch_size*p
+        batch_size = test_rep.shape[0]
+        for layer in range(layer_cnt):
+            grad_layer = full_rep_split[layer]
+            # length = full train data size
+            length = grad_layer.shape[0]
+            # Split gradients into smaller chunks
+            grad_batches = grad_layer.split(batch_size, dim=0)
+            running_contributions = 0
```
jiaqima, line 700: running_transformation

jiaqima: Also, this quantity will actually become a tensor. This initial value may be confusing. Let the initial value be a torch.zeros() with the same shape, type, and device as the original test_rep?

dattri/algorithm/influence_function.py
```diff
@@ -620,35 +649,94 @@ def transform_test_rep(
...
+                    information matrix in each layer.
+
+            Returns:
+                torch.Tensor: A tensor corresponding to transformed test representation.
+            """
+            grad = grad.unsqueeze(-1)
+            coef = (v @ grad) / (regularization + torch.norm(grad) ** 2)
+            return (v - coef @ grad.T) / regularization
+
+        regularization = self.transformation_kwargs["regularization"]
```
jiaqima, line 684: let's directly store the value in self.regularization at initialization and get rid of self. transformation_kwargs, since that's the only elements in this transformation_kwargs

jiaqima (changes requested):
Thanks for the update. Please see comments for further improvement.

dattri/func/fisher.py
jiaqima: I think you could update the test in this PR as well.

charles-pyj: Sure. Removing all tests of ifvp_datainf

dattri/algorithm/influence_function.py
```diff
@@ -620,35 +652,95 @@ def transform_test_rep(
...
+        layer_cnt = len(cached_train_rep_layers)
+        transformed_test_rep_layers = []
+        # Use test batch size as intermediate batch size
+        # Peak memory usage: max(train_batch_size,test_batch_size)*test_batch_size*p
+        batch_size = test_rep.shape[0]
+        for layer in range(layer_cnt):
+            grad_layer = cached_train_rep_layers[layer]
+            # length = full train data size
+            length = grad_layer.shape[0]
+            # Split gradients into smaller chunks
```
jiaqima, line 701: chunks -> batches

dattri/algorithm/influence_function.py
```diff
@@ -620,35 +652,95 @@ def transform_test_rep(
...
+            for batch in grad_batches:
+                reg = 0.1 if regularization is None else regularization
+                contribution = torch.func.vmap(
+                    lambda grad, layer=layer, reg=reg: _transform_single_test_rep(
+                    test_rep_layers[layer],
+                    grad,
+                    reg,
+                ),
+                )(batch)
+                # Sum up the contributions by chunk and average at the end
```
jiaqima, line 714: accumulate the batches

jiaqima (changes requested):
One last thing: given the substantial implementation for this attributor is moved to the attributor class (instead of an ihvp function), could you add a few tests in test/algorithm/test_influence_function.py to test the implementation of this attributor? Ideally, we may want to test the both the attributor as a whole and the transform_test_rep function

dattri/algorithm/influence_function.py
```diff
@@ -586,6 +586,7 @@ def __init__(
         task: AttributionTask,
         device: Optional[str] = "cpu",
         regularization: float = 0.0,
+        full_data_ratio: float = 1.0,
```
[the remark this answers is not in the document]
charles-pyj: Fixed.

dattri/algorithm/influence_function.py
```diff
@@ -597,18 +598,46 @@ def __init__(
...
+            for full_data_ in full_train_dataloader:
+                sample_size = math.ceil(full_data_[0].shape[0] * self.full_data_ratio)
+                sampled_data = tuple(data.to(self.device)[:sample_size].unsqueeze(0)
+                    for data in full_data_)
+                sample_full_data_rep = self.generate_train_rep(
+                    ckpt_idx=checkpoint_idx,
+                    data=sampled_data,
+                )
+                _cached_train_reps_list.append(sample_full_data_rep)
+            self._cached_train_reps[f"ckpt_{checkpoint_idx}"] = torch.cat(
```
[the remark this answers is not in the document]
charles-pyj: Fixed.

dattri/algorithm/influence_function.py
charles-pyj: test_rep_layers seem to be a more informative name. Solved.

dattri/algorithm/influence_function.py
charles-pyj: Fixed.

dattri/algorithm/influence_function.py
charles-pyj: Removed kwargs and added new attribute regularization.

dattri/algorithm/influence_function.py
charles-pyj: Initialized to be the same size and device as layer-wise test reps

dattri/algorithm/influence_function.py
charles-pyj: Changed name to trainsformed_test_rep_layers.

charles-pyj:
> One last thing: given the substantial implementation for this attributor is moved to the attributor class (instead of an ihvp function), could you add a few tests in test/algorithm/test_influence_function.py to test the implementation of this attributor? Ideally, we may want to test the both the attributor as a whole and the transform_test_rep function

Just to clarify, the test_influence_function.py does have a simple test of datainf that only tests successful running. We would like a more extended testing of the results as well right?

jiaqima:
> > One last thing: given the substantial implementation for this attributor is moved to the attributor class (instead of an ihvp function), could you add a few tests in test/algorithm/test_influence_function.py to test the implementation of this attributor? Ideally, we may want to test the both the attributor as a whole and the transform_test_rep function
> 
> Just to clarify, the test_influence_function.py does have a simple test of datainf that only tests successful running. We would like a more extended testing of the results as well right?

Yes. I think it would be better to test the intermediate function, transform_test_rep, as well. For example, you could craft some random initial test_rep and set the train gradients to cache to be very small. Then you could pull out the cached train gradients, and calculate the transformed test_rep by directly plug in the formula. Then you compare the calculated results with the output of the transform_test_rep.

dattri/algorithm/influence_function.py
charles-pyj: Updated comments

dattri/algorithm/influence_function.py
charles-pyj: Updated comments

charles-pyj:
Got it! The test is added for transform test representation.

jiaqima (approved):
LGTM

PR #126, opened 2024-08-17
[dattri.algorithm] Fix misalignment in KNN shapley

sx-liu:
## Description

In the previous version of `KNNShalpeyAttributor.attribute`, the test labels are misaligned when there are multiple batches in the test loader. This PR fixes this bug.


TheaperDeng (approved):
LGTM

Issue #29, opened 2024-04-21
dattri.metrics.metrics is  a bit redundant


jiaqima:
Currently, we have our evaluation metrics included in dattri/metrics/metrics.py, which leads to the API `dattri.metrics.metrics.xxx`. This is a bit redundant. We may want to remove one 'metrics' in this path.

Issue #26, opened 2024-04-17
module name consistency 


TheaperDeng:
for consistency, we should either have dattri.metric, dattri.benchmark, dattri.algorithm, OR dattri.metrics, dattri.benchmarks, dattri.algorithms

jiaqima:
We could remove "s" under dattri, since it's a bit weird to have dattri.funcs

PR #128, opened 2024-09-03
[dattri experiments] benchmark script for mnist, cifar and shakespeare

TheaperDeng:
## Description

A script to run benchmark in one line. And upload the checkpoints and groundtruth for

- [x] Mnist + LR
- [x] Mnist + MLP
- [x] Cifar2 + ResNet9
- [ ] MAESTRO + MT (leave for @tingwl0122 )
- [x] Shakespeare + nanoGPT

### Summary of changes
1. Please check https://github.com/TheaperDeng/dattri-public/tree/experiments-update/experiments for the usage of the benchmark script.
2. Remove other files in `experiments/` folder. One of them (`mnist_lr_brittleness.py`) is moved to `example/` and the others are moved since the functionality is covered by the new script.


experiments/benchmark_result.py
```diff
@@ -0,0 +1,335 @@
...
+            p = torch.exp(-loss(yhat, label_t.long()))
+            return p
+
+    if args.method == "RPS":
+
+        def loss_rps(pre_activation_list, label_list):
+            loss_fn = nn.CrossEntropyLoss()
+            return loss_fn(pre_activation_list, label_list)
+
+        model = model_details["model"].to(args.device)
```
jiaqima, line 205: are these two lines redundant?

experiments/benchmark_result.py
```diff
@@ -0,0 +1,335 @@
...
+                device=args.device,
+                **ihvp_config,
+            )
+            attributor.cache(train_loader_cache)
+            with torch.no_grad():
+                score = attributor.attribute(train_loader, test_loader)
+
+            # compute metrics
+            if args.metric == "lds":
+                score = -score
```
jiaqima, line 232: why?

jiaqima: I think it's better to make the API of all metrics consistent

jiaqima (approved):
Others LGTM

jiaqima:
What's the difference between example and experiments?

tingwl0122:
> What's the difference between example and experiments?

I do think they can be merged into only one folder. They are just different experiments while `experiments` are directly related to the ones provided in the paper.

TheaperDeng:
> What's the difference between example and experiments?

I think this as a user experience direction.

`experiments/` contains the one-line script that could **reproduce the result we posted on our paper**, but it's not a good reference taken by users to write their own code. Because it's long (due to the support of multiple different attributors and the hyperparameter searching) and may take more time than a normal example.

`examples/` contains some bite-sized scripts that can help our users to understand how to use our toolkit. They only do one thing in a script, with small dataset to make sure everyone can run it in a short time.

Thats why I say " You may find some easier and clearer examples in /example folder." in the [readme](https://github.com/TheaperDeng/dattri-public/tree/experiments-update/experiments).

experiments/benchmark_result.py
TheaperDeng: Oh, this should be the LDS groundtruth calculation problem, since the LDS groundtruth is calculated by `loss` for now, it should be `-loss` since LDS using L_{+z} and LOO using L_{-z} (where z is some training sample).

I guess the best solution is to make all the LDS groundtruth we upload to be negative of the current version?

TheaperDeng: > Or raise and merge a PR about the API change first and then update this PR?
> 
> We may also address issues #26 and #29 first.

Thanks! If we decided to change the groundtruth to make this API consistent, we may fix these two issues together in a seperated PR.

tingwl0122:
One minor thing, after `load_benchmark` and scripts supports all the settings, we can probably copy the readme in https://github.com/TheaperDeng/dattri-public/tree/experiments-update/experiments  under `experiments`?

experiments/benchmark_result.py
TheaperDeng: Fixed the groundtruth.

jiaqima:
> > What's the difference between example and experiments?
> 
> I think this as a user experience direction.
> 
> `experiments/` contains the one-line script that could **reproduce the result we posted on our paper**, but it's not a good reference taken by users to write their own code. Because it's long (due to the support of multiple different attributors and the hyperparameter searching) and may take more time than a normal example.
> 
> `examples/` contains **some bite-sized scripts that can help our users to understand how to use our toolkit**. They only do one thing in a script, with small dataset to make sure everyone can run it in a short time.
> 
> Thats why I say " You may find some easier and clearer examples in /example folder." in the [readme](https://github.com/TheaperDeng/dattri-public/tree/experiments-update/experiments).

minor: /example -> /examples

TheaperDeng:
> > > What's the difference between example and experiments?
> > 
> > 
> > I think this as a user experience direction.
> > `experiments/` contains the one-line script that could **reproduce the result we posted on our paper**, but it's not a good reference taken by users to write their own code. Because it's long (due to the support of multiple different attributors and the hyperparameter searching) and may take more time than a normal example.
> > `examples/` contains **some bite-sized scripts that can help our users to understand how to use our toolkit**. They only do one thing in a script, with small dataset to make sure everyone can run it in a short time.
> > Thats why I say " You may find some easier and clearer examples in /example folder." in the [readme](https://github.com/TheaperDeng/dattri-public/tree/experiments-update/experiments).
> 
> minor: /example -> /examples

Thanks. I will change it in next PR which handles the CI/CD of examples.

PR #130, opened 2024-09-04
[DO NOT MERGE] Reference refactoring of BaseInnerProductAttributor

jiaqima:
## Description

This PR is a reference implementation to refactor the BaseInnerProductAttributor, which is NOT intended to be merged (and has not been tested). As an example, the Arnoldi attributor is re-implemented to demonstrate how the new API of BaseInnerProductAttributor can be used.

A few major changes in BaseInnerProductAttributor:
- transformation could now be applied to both train data representations/gradients and test data representations/gradients
- the batching over the full training data for transformation was original dealt inside `.attribute()` with a simple average, which lacks enough flexibility; it is now moved into the transform functions
- `.cache()` could do more than just storing the full training dataloader. For example, the Arnoldi projector is now processed and cached in transformation_on_query in a hacky way. In this PR, this is moved to .cache()
- removed transformation_kwargs and added an __init__ for the subclass. It's better to specify these attributor hyperparameters explicitly than relying on a kwargs

### DIFF
--- a/dattri/algorithm/base.py
+++ b/dattri/algorithm/base.py
@@ -41,7 +41,7 @@ class BaseAttributor(ABC):
 
         Args:
             full_train_dataloader (torch.utils.data.DataLoader): The dataloader for
-                the training data.
+                the full training data.
 
         Returns:
             None.
@@ -83,7 +83,7 @@ class BaseInnerProductAttributor(BaseAttributor):
             transformation_kwargs (Optional[Dict[str, Any]]): The keyword arguments for
                 the transformation function. More specifically, it will be stored in
                 the `transformation_kwargs` attribute and be used by some customized
-                member functions, e.g., `transformation_on_query`, where the
+                member functions, e.g., `transform_test_rep`, where the
                 transformation such as hessian matrix or Fisher Information matrix is
                 calculated.
             device (str): The device to run the attributor.
@@ -91,7 +91,6 @@ class BaseInnerProductAttributor(BaseAttributor):
         self.task = task
         self.transformation_kwargs = transformation_kwargs or {}
         self.device = device
-        self.index = 0
 
     def _set_test_data(self, dataloader: torch.utils.data.DataLoader) -> None:
         """Set test dataloader.
@@ -99,7 +98,7 @@ class BaseInnerProductAttributor(BaseAttributor):
         Args:
             dataloader (DataLoader): The dataloader for test samples to be attributed.
         """
-        # This function may be overrided by the subclass
+        # This function may be overridden by the subclass
         self.test_dataloader = dataloader
 
     def _set_train_data(self, dataloader: torch.utils.data.DataLoader) -> None:
@@ -108,7 +107,7 @@ class BaseInnerProductAttributor(BaseAttributor):
         Args:
             dataloader (DataLoader): The dataloader for train samples to be attributed.
         """
-        # This function may be overrided by the subclass
+        # This function may be overridden by the subclass
         self.train_dataloader = dataloader
 
     def _set_full_train_data(self, dataloader: torch.utils.data.DataLoader) -> None:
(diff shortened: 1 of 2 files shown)


dattri/algorithm/base.py
```diff
@@ -117,104 +108,127 @@ def _set_full_train_data(self, dataloader: torch.utils.data.DataLoader) -> None:
...
     ) -> torch.Tensor:
-        """Calculating the query based on the train data.
+        """Generate the initial representations of the train data.
 
-        Inner product attributor calculates the inner product between the
-        train query and the transformation of the test query. This function
-        calculates the train query based on the train data.
+        Inner product attributor calculates the inner product between the (transformed)
+        train representations and test representations. This function generates the
+        initial train representations.
```
TheaperDeng, line 150: The default implementation calculates the gradient of the train loss with respect to the parameter. Subclasses may override this function to calculate something else.

TheaperDeng (commented):
I am OK with these changes and will make another PR to implement them with solid test.

jiaqima:
@TheaperDeng, one thing I'm not sure is if we want to set the default transform functions for both train and test as identical transformation? As in many cases, only test rep needs to be transformed.

TheaperDeng:
> @TheaperDeng, one thing I'm not sure is if we want to set the default transform functions for both train and test as identical transformation? As in many cases, only test rep needs to be transformed.

It happens that I am working on this. I think one of the possible way is that, we just make `transform_test_rep` to be `abstractmethod`, that is users always need to implement this one. and for `transform_train_rep`, we can have a default implementation that returns the train rep directly.

jiaqima:
> > @TheaperDeng, one thing I'm not sure is if we want to set the default transform functions for both train and test as identical transformation? As in many cases, only test rep needs to be transformed.
> 
> It happens that I am working on this. I think one of the possible way is that, we just make `transform_test_rep` to be `abstractmethod`, that is users always need to implement this one. and for `transform_train_rep`, we can have a default implementation that returns the train rep directly.
> 
> Or, we directly set them both to be defaultly return the rep directly. Users can override them when design their TDA methods.

I think the latter is Ok. Users implementing new attributors should be a bit more tolerant than users who just use the attributors, and they don't need an abstract method to remind them this is something they should override.

TheaperDeng:
> > > @TheaperDeng, one thing I'm not sure is if we want to set the default transform functions for both train and test as identical transformation? As in many cases, only test rep needs to be transformed.
> > 
> > 
> > It happens that I am working on this. I think one of the possible way is that, we just make `transform_test_rep` to be `abstractmethod`, that is users always need to implement this one. and for `transform_train_rep`, we can have a default implementation that returns the train rep directly.
> > Or, we directly set them both to be defaultly return the rep directly. Users can override them when design their TDA methods.
> 
> I think the latter is Ok. Users implementing new attributors should be a bit more tolerant than users who just use the attributors, and they don't need an abstract method to remind them this is something they should override.

Yeah, I think so. And this makes the innerproduct IF can be directly instantiated and works like Grad-Dot, just like its name.

jiaqima:
@tingwl0122 could you please take a look at the new implementation of the Arnoldi attributor and see if there is anything wrong?

tingwl0122:
> @tingwl0122 could you please take a look at the new implementation of the Arnoldi attributor and see if there is anything wrong


I think the overall structure looks good to me. (Moving projection matrix computation to cache has been listed as one todo) Another thing is that the original Arnoldi paper actually used only a random batch of the full dataloader to do the algorithm, which I didn't include before. We can do something on the `data_target_pair` before feeding it into `func` to achieve this now. (also adding an additional param to control this random batch's size)

tingwl0122:
> > @tingwl0122 could you please take a look at the new implementation of the Arnoldi attributor and see if there is anything wrong
> 
> I think the overall structure looks good to me. (Moving projection matrix computation to cache has been listed as one todo) Another thing is that the original Arnoldi paper actually used only a random batch of the full dataloader to do the algorithm, which I didn't include before. We can do something on the `data_target_pair` before feeding it into `func` to achieve this now. (also adding an additional param to control this random batch's size)

another minor thing is that we will need to force users to use `.cache` before attributing. We can probably raise some errors for this.

jiaqima:
> > > @tingwl0122 could you please take a look at the new implementation of the Arnoldi attributor and see if there is anything wrong
> > 
> > 
> > I think the overall structure looks good to me. (Moving projection matrix computation to cache has been listed as one todo) Another thing is that the original Arnoldi paper actually used only a random batch of the full dataloader to do the algorithm, which I didn't include before. We can do something on the `data_target_pair` before feeding it into `func` to achieve this now. (also adding an additional param to control this random batch's size)
> 
> another minor thing is that we will need to force users to use `.cache` before attributing. We can probably raise some errors for this.

Sounds great. Thanks for the notes. I'll add them later.

TheaperDeng:
> > > > @tingwl0122 could you please take a look at the new implementation of the Arnoldi attributor and see if there is anything wrong
> > > 
> > > 
> > > I think the overall structure looks good to me. (Moving projection matrix computation to cache has been listed as one todo) Another thing is that the original Arnoldi paper actually used only a random batch of the full dataloader to do the algorithm, which I didn't include before. We can do something on the `data_target_pair` before feeding it into `func` to achieve this now. (also adding an additional param to control this random batch's size)
> > 
> > 
> > another minor thing is that we will need to force users to use `.cache` before attributing. We can probably raise some errors for this.
> 
> Sounds great. Thanks for the notes. I'll add them later.

Thanks @tingwl0122 @jiaqima I think I can directly make the changes in my implementation #133

jiaqima:
Closing this PR in favor of #133. Any follow-up discussion could be done there.

PR #132, opened 2024-09-05
[dattri.metrics] change the shape of score in `lds` to improve consistency

TheaperDeng:
### 1. Motivation and Context
Currently, we have all the attributor and `loo_corr` accept score in shape of (num_train_samples, num_test_samples). It would be strange if we have `lds` accept (num_test_samples, num_train_samples). 

### 2. Summary of the change
1. Change the shape of `score` in `dattri.metrics.metrics.lds` to be (num_train_samples, num_test_samples).
2. Add a cpu transformation since the calculation will happen in cpu for the correlation calculation, this avoid some possible error in users' code if they do not change the device explicitly.


tingwl0122:
LGTM

PR #133, opened 2024-09-08
[dattri.algorithm] Reference refactoring of BaseInnerProductAttributor

TheaperDeng:
### 1. Motivation and Context
Revise the `BaseInnerProductAttributor` following the spirit of https://github.com/TRAIS-Lab/dattri/pull/130

### 2. Summary of the change

- transformation could now be applied to both train data representations/gradients and test data representations/gradients
- the batching over the full training data for transformation was original dealt inside .attribute() with a simple average, which lacks enough flexibility; it is now moved into the transform functions
- .cache() could do more than just storing the full training dataloader. For example, the Arnoldi projector is now processed in
- removed transformation_kwargs and added an init for the subclass. It's better to specify these attributor hyperparameters explicitly than relying on a kwargs

### DIFF
--- a/dattri/algorithm/base.py
+++ b/dattri/algorithm/base.py
@@ -6,7 +6,7 @@ from abc import ABC, abstractmethod
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    from typing import Any, Dict, Optional, Tuple
+    from typing import Optional, Tuple
 
     import torch
     from torch.utils.data import DataLoader
@@ -41,7 +41,7 @@ class BaseAttributor(ABC):
 
         Args:
             full_train_dataloader (torch.utils.data.DataLoader): The dataloader for
-                the training data.
+                the full training data.
 
         Returns:
             None.
@@ -73,25 +73,16 @@ class BaseInnerProductAttributor(BaseAttributor):
         self,
         task: AttributionTask,
         device: Optional[str] = "cpu",
-        **transformation_kwargs: Dict[str, Any],
     ) -> None:
         """Initialize the attributor.
 
         Args:
             task (AttributionTask): The task to be attributed. The task should
                 be an instance of `AttributionTask`.
-            transformation_kwargs (Optional[Dict[str, Any]]): The keyword arguments for
-                the transformation function. More specifically, it will be stored in
-                the `transformation_kwargs` attribute and be used by some customized
-                member functions, e.g., `transformation_on_query`, where the
-                transformation such as hessian matrix or Fisher Information matrix is
-                calculated.
             device (str): The device to run the attributor.
         """
         self.task = task
-        self.transformation_kwargs = transformation_kwargs or {}
         self.device = device
-        self.index = 0
 
     def _set_test_data(self, dataloader: torch.utils.data.DataLoader) -> None:
         """Set test dataloader.
@@ -99,7 +90,7 @@ class BaseInnerProductAttributor(BaseAttributor):
         Args:
             dataloader (DataLoader): The dataloader for test samples to be attributed.
         """
-        # This function may be overrided by the subclass
+        # This function may be overridden by the subclass
         self.test_dataloader = dataloader
 
     def _set_train_data(self, dataloader: torch.utils.data.DataLoader) -> None:
@@ -108,7 +99,7 @@ class BaseInnerProductAttributor(BaseAttributor):
         Args:
             dataloader (DataLoader): The dataloader for train samples to be attributed.
         """
-        # This function may be overrided by the subclass
+        # This function may be overridden by the subclass
         self.train_dataloader = dataloader
 
     def _set_full_train_data(self, dataloader: torch.utils.data.DataLoader) -> None:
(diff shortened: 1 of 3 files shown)


TheaperDeng:
@jiaqima @tingwl0122 please have a look at this PR. The update summary is in the description.

There might be some inconsistent in the docstring and I will give it a pass later today :)

dattri/algorithm/influence_function.py
```diff
@@ -222,169 +225,362 @@ def attribute(
...
-            torch.Tensor: The transformation on the query. Normally it is a 2-d
-                dimensional tensor with the shape of (batchsize, transformed_dimension).
+        Args:
+            full_train_dataloader (DataLoader): The dataloader with full training data.
         """
+        self.full_train_dataloader = full_train_dataloader
+        self.arnoldi_projectors = []
+
+        # Assuming that full_train_dataloader has only one batch
+        # TODO: support multiple batches
```
jiaqima, line 427: Remove this comment since it has been addressed

dattri/algorithm/influence_function.py
```diff
@@ -222,169 +225,362 @@ def attribute(
...
+
+        Raises:
+            ValueError: If the Arnoldi projector has not been cached.
+        """
+        if not hasattr(self, "arnoldi_projectors"):
+            error_msg = "The Arnoldi projector has not been cached.\
+                         Please call cache() first."
+            raise ValueError(error_msg)
+
+        return self.arnoldi_projectors[ckpt_idx](test_rep).detach()
```
jiaqima, line 498: I think we also need to transform train_rep? @tingwl0122

jiaqima (approved):
Overall LGTM. Left some minor comments.

dattri/algorithm/influence_function.py
tingwl0122: for the arnoldi projector written in `dattri/func/projection.py`,
the projection is defined as (eigvecs * inverse of eigvals * eigvecs.T) * (test or train feature),
so we only need to apply it to one of test or train;

while in the original paper, it define the projection as (eigvec.T) * (test or train feature),
so it needs to apply projection to both train and test, along with the inverse eigvals in the middle.

dattri/algorithm/base.py
```diff
@@ -117,104 +109,135 @@ def _set_full_train_data(self, dataloader: torch.utils.data.DataLoader) -> None:
...
+            torch.Tensor: The transformed train representations. Typically,
+                it is a 2-d dimensional tensor with the shape of
+                (batch_size, transformed_dimension).
         """
+        return train_rep
 
-    def cache(self, full_train_dataloader: DataLoader) -> None:
-        """Cache the dataset for inverse hessian calculation.
+    def cache(self, full_train_dataloader: torch.utils.data.DataLoader) -> None:
+        """Cache the full training dataloader.
```
tingwl0122, line 230: or mention some pre-computation  can also be cached?
since it can be more things

dattri/algorithm/influence_function.py
tingwl0122: For the original paper, the projection is defined as "eigvec.T * train/test feature";
while we defined it as "eigvec * inverse eigvals * eigvec.T * train/test feature".
Thus, we only need to apply it to either one of the train or test feature

README.md
```diff
@@ -162,6 +180,7 @@ model = activate_dropout(model, ["dropout1", "dropout2"], dropout_prob=0.2)
 |        |   [Grad-Cos](https://arxiv.org/abs/2102.05262)  |
 |   [RPS](https://arxiv.org/abs/1811.09720)  |   [RPS-L2](https://arxiv.org/abs/1811.09720)   |
 |  [TRAK](https://arxiv.org/abs/2303.14186)  |       [TRAK](https://arxiv.org/abs/2303.14186)       |
+|  [Shapley Value](https://arxiv.org/abs/1904.02868)  |       [KNN-Shapley](https://dl.acm.org/doi/10.14778/3342263.3342637)       |
```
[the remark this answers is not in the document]
TheaperDeng: No we don't have it. Just have KNN-Shapley in this family.

dattri/algorithm/base.py
```diff
@@ -117,104 +109,135 @@ def _set_full_train_data(self, dataloader: torch.utils.data.DataLoader) -> None:
...
-            data (Tuple[Tensor]): The test data. Normally this is a tuple
-                of input data and target data, the number of items in the
-                tuple should be aligned in the target function. The tensors'
-                shape follows (1, batchsize, ...).
+            ckpt_idx (int): The index of the model checkpoints. This index
+                is used for ensembling different trained model checkpoints.
+            data (Tuple[Tensor]): The test data. Typically, this is a tuple
+                of input data and target data but the number of items in this
+                tuple should align with the corresponding argument in the
+                target function. The tensors' shape follows (1, batch_size, ...).
```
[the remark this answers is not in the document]
TheaperDeng: I guess I can open another PR to fix this (at least try to make them aligned as much as possible)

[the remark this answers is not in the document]
tingwl0122: probably not a really big deal for now since users can follow our examples, but it might cause some problems since those unsqueeze stuff might not be trivial or default for users.

README.md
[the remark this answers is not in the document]
tingwl0122: oh it is the family name, not the method itself.
My bad, resolved

dattri/algorithm/base.py
```diff
@@ -117,104 +109,135 @@ def _set_full_train_data(self, dataloader: torch.utils.data.DataLoader) -> None:
...
         """
-        # This function may be overrided by the subclass
+        # This function may be overridden by the subclass
         self.full_train_dataloader = dataloader
 
-    def generate_test_query(
+    def generate_test_rep(
         self,
-        index: int,
+        ckpt_idx: int,
```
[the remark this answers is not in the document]
TheaperDeng: I think this should be OK? Since no one (even the TDA developers) need to explicitly write any code calling `generate_test_rep` / `transform_test_rep` / `generate_train_rep` / `transform_train_rep`. Then I guess this is okay to leave it as it is.

dattri/algorithm/influence_function.py
jiaqima: In this case, we are storing a num_parameter x num_parameter matrix in the projector?

It's probably more memory efficient to define the projection as (sqrt(1/eigvals) * eigvec.T)[topk_index]?

tingwl0122: I think this is because we regard this as a inverse-hessian multiplier before.

jiaqima: Yeah, that's likely the reason. Now that we treat it as a projection, could you please raise a PR to make this more efficient?

tingwl0122: Just raised a PR for this

PR #135, opened 2024-09-11
[dattri.func] Re-implementation of Arnoldi projector

tingwl0122:
### 1. Motivation and Context

Re-define the arnoldi projector as 1.0 / sqrt(eigvals) * eigvecs.T * features

### 2. Summary of the change
- fix the `arnoldi_project`
- add `transform_train_rep` inside arnoldi attributor
- fix a slight typo found in `dattri/benchmark/datasets/mnist/mnist_mlp`

### DIFF
--- a/dattri/func/projection.py
+++ b/dattri/func/projection.py
@@ -23,6 +23,8 @@ if TYPE_CHECKING:
     from collections.abc import Callable
     from typing import Dict, List, Union
 
+import warnings
+
 import numpy as np
 import torch
 from torch import Tensor
@@ -753,6 +755,18 @@ class ArnoldiProjector(AbstractProjector):
         # proj dim is the value of k
         self.eigvals, self.eigvecs = self._distill(appr_mat, proj, self.proj_dim)
 
+        # prevent from negative eigvals
+        if self.proj_dim > torch.sum(self.eigvals > 0):
+            # adjust proj_dim
+            self.proj_dim = torch.sum(self.eigvals > 0).item()
+            warnings.warn(
+                "Encountered many negative eigenvalues and `proj_dim` is greater"
+                " than the number of positive eigenvalues. Automatically adjusting"
+                " `proj_dim` to the number of positive eigenvalues. Please consider"
+                " increasing `regularization` to reduce negative eigenvalues.",
+                stacklevel=1,
+            )
+
     def project(self, features: Union[dict, Tensor]) -> Tensor:
         """Performs the random projection on the feature matrix.
 
@@ -770,7 +784,6 @@ class ArnoldiProjector(AbstractProjector):
         if self.eigvals is None or self.eigvecs is None:
             self.get_eigenspace()
 
-        self.eigvals = (self.eigvals).to(torch.complex64)
         return features @ self.eigvecs.T * (1.0 / torch.sqrt(self.eigvals.unsqueeze(0)))
 
     def free_memory(self) -> None:
--- a/dattri/func/projection.py
+++ b/dattri/func/projection.py
@@ -769,10 +769,7 @@ class ArnoldiProjector(AbstractProjector):
         # have not compute the eigen space yet
         if self.eigvals is None or self.eigvecs is None:
             self.get_eigenspace()
-
-        return (
-            (features @ self.eigvecs.T) * 1.0 / self.eigvals.unsqueeze(0)
-        ) @ self.eigvecs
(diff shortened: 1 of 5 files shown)


tingwl0122:
Tested on influence_function_noisy_label task
```
0                        0                        
100                      69                       
200                      100                      
300                      106                      
400                      108                      
500                      108                      
600                      108                      
700                      108                      
800                      108                      
900                      108  
```

TheaperDeng (commented):
LGTM, the unittest needs to be changed accordingly.

TheaperDeng:
We may remove `dattri.func.hessian.ihvp_arnoldi` and `dattri.func.hessian.ihvp_at_x_arnoldi` later? Currently only the old IFAttributor is using it and we are defining arnoldi as a projector now. @jiaqima

jiaqima:
> We may remove `dattri.func.hessian.ihvp_arnoldi` and `dattri.func.hessian.ihvp_at_x_arnoldi` later? Currently only the old IFAttributor is using it and we are defining arnoldi as a projector now. @jiaqima  

Good point! I think we could remove them when we get rid of the old IFAttributor.

tingwl0122:
Another point is that we might have to handle negative eigen values if we want to do this sqrt thing.
We can definitely cope with complex numbers but it might be a bit messy.

jiaqima:
> Another point is that we might have to handle negative eigen values if we want to do this sqrt thing. We can definitely cope with complex numbers but it might be a bit messy.

How about we project to min(proj_dim, # positive eigvals)? If # positive eigvals < proj_dim, throw out a warning about that. In this case, we are always projecting to eig space with positive eig vals. I believe in practice, # positive eigvals should usually be larger than proj_dim?

tingwl0122:
> > Another point is that we might have to handle negative eigen values if we want to do this sqrt thing. We can definitely cope with complex numbers but it might be a bit messy.
> 
> How about we project to min(proj_dim, # positive eigvals)? If # positive eigvals < proj_dim, throw out a warning about that. In this case, we are always projecting to eig space with positive eig vals. I believe in practice, # positive eigvals should usually be larger than proj_dim?

I am pretty unsure about this, I indeed use quite large proj_dim (close to param size actually) during some experiments during dattri rebuttal.
now the newest version can handle complex numbers with minimal modification.

jiaqima:
> > > Another point is that we might have to handle negative eigen values if we want to do this sqrt thing. We can definitely cope with complex numbers but it might be a bit messy.
> > 
> > 
> > How about we project to min(proj_dim, # positive eigvals)? If # positive eigvals < proj_dim, throw out a warning about that. In this case, we are always projecting to eig space with positive eig vals. I believe in practice, # positive eigvals should usually be larger than proj_dim?
> 
> I am pretty unsure about this, I indeed use quite large proj_dim (close to param size actually) during some experiments during dattri rebuttal. now the newest version can handle complex numbers with minimal modification.

I'm afraid that this would mess up some subsequent computation as everything is complex type after the projection?

tingwl0122:
> > > > Another point is that we might have to handle negative eigen values if we want to do this sqrt thing. We can definitely cope with complex numbers but it might be a bit messy.
> > > 
> > > 
> > > How about we project to min(proj_dim, # positive eigvals)? If # positive eigvals < proj_dim, throw out a warning about that. In this case, we are always projecting to eig space with positive eig vals. I believe in practice, # positive eigvals should usually be larger than proj_dim?
> > 
> > 
> > I am pretty unsure about this, I indeed use quite large proj_dim (close to param size actually) during some experiments during dattri rebuttal. now the newest version can handle complex numbers with minimal modification.
> 
> I'm afraid that this would mess up some subsequent computation as everything is complex type after the projection?

we just need to convert it back to float since there will be no imaginary part after inner product. So I guess as long as we are still under `BaseInnerProductAttributor`, then we should be ok.

tingwl0122:
Now we will throw a warning if `proj_dim` is larger than the number of positive eigvals and will automatically adjust `proj_dim` to be equal to this number if that happens. Test case will be adjusted accordingly

dattri/func/projection.py
```diff
@@ -753,6 +755,18 @@ def get_eigenspace(self) -> None:
         # proj dim is the value of k
         self.eigvals, self.eigvecs = self._distill(appr_mat, proj, self.proj_dim)
 
+        # prevent from negative eigvals
+        if self.proj_dim > torch.sum(self.eigvals > 0):
```
jiaqima, line 759: Maybe change "> 0" to "> self.tol**2" to improve numerical stability

tingwl0122: Sure

PR #140, opened 2024-09-14
[dattri.algorithm] relax the requirement to the `loss_func`.

TheaperDeng:
### 1. Motivation and Context

Currently, IF Attributors (CG, Explicit, Arnoldi) require the second parameter (data) of `loss_func` to be named as "data_target_pair". This PR relax this requirement.

### 2. Summary of the change

Inspect the loss function's signature and use it in the implementation of IF


jiaqima (approved):
LGTM

PR #143, opened 2024-09-17
Refactor the implementation of EKFAC

sx-liu:
### 1. Motivation and Context

The current implementation of EK-FAC is out-dated. Refactor the EK-FAC attribution to follow the format of other IF attribution, as well as update the hook mechanism to avoid redundency.

### 2. Summary of the change

1. Refactor EK-FAC base functions, add `estimate_covariance`, `estimate_eigenvector` and `estimate_lambda` for direct use by the users.
2. Remove `ifvp_at_x_ekfac`, add `IFAttributorEKFAC` instead for attribution.
3. Remove `MLPCache` class and `manual_cache_forward` function, and use `torch.register_forward_hook` instead.

- [ ] ...

### DIFF
--- a/dattri/func/fisher.py
+++ b/dattri/func/fisher.py
@@ -321,13 +321,13 @@ def _random_batch_iterator(
         )
 
 
-def _estimate_covariance(
+def _update_covariance(
     curr_estimate: List[List[Tuple[torch.Tensor]]],
     mlp_cache: List[MLPCache],
     total_samples: int,
     mask: torch.Tensor,
 ) -> List[List[Tuple[torch.Tensor]]]:
-    """Estimate the 'covariance' matrices S and A in EK-FAC IFVP.
+    """Update the running estimation of the 'covariance' matrices S and A in EK-FAC IFVP.
 
     Args:
         curr_estimate (List[List[Tuple[torch.Tensor]]]): A list of lists of tuples
@@ -373,7 +373,7 @@ def _estimate_covariance(
     return curr_estimate
 
 
-def _estimate_lambda(
+def _update_lambda(
     curr_estimate: List[List[torch.Tensor]],
     mlp_cache: List[MLPCache],
     cached_q: List[List[Tuple[torch.Tensor]]],
@@ -381,7 +381,7 @@ def _estimate_lambda(
     mask: torch.Tensor,
     max_steps_for_vec: int = 10,
 ) -> List[List[torch.Tensor]]:
-    """Estimate the corrected eigenvalues in EK-FAC IFVP.
+    """Update the running estimation of the corrected eigenvalues in EK-FAC IFVP.
 
     Args:
         curr_estimate (List[List[torch.Tensor]]): A list of lists of tensors,
@@ -465,13 +465,98 @@ def _estimate_lambda(
     return curr_estimate
 
 
+def estimate_covariance(
+    func: Callable,
+    dataloader: torch.utils.data.DataLoader,
+    mlp_cache: List[MLPCache],
+    max_iter: int,
+) -> List[List[Tuple[torch.Tensor]]]:
+    covariances = [[] for _ in range(len(mlp_cache))]
+    total_samples = 0  # record total number of samples
+    for i, batch in enumerate(dataloader):
+        # Forward pass
+        func_output = func(*batch)
+        losses, mask = (
+            func_output if isinstance(func_output, tuple) else func_output,
+            torch.tensor(1.0),
+        )
+
+        for loss in losses:
+            # Backward pass
+            loss.backward(retain_graph=True)
+
+        with torch.no_grad():
+            # Estimate covariance
+            cov_matrices = _update_covariance(
+                cov_matrices,
+                mlp_cache,
+                total_samples,
+                mask,
+            )
+
+        total_samples += int(mask.sum())
+        if i == max_iter - 1:
+            break
+
+    return covariances
+
+
+def estimate_eigenvector(
+    covariances: List[List[Tuple[torch.Tensor]]],
+    mlp_cache: List[MLPCache],
+) -> List[List[Tuple[torch.Tensor]]]:
+    cached_q = [[] for _ in range(len(mlp_cache))]
+    for layer_q, layer_cov in zip(cached_q, covariances):
+        for cov_a, cov_s in layer_cov:
+            _, q_a = torch.linalg.eigh(cov_a, UPLO="U")
+            _, q_s = torch.linalg.eigh(cov_s, UPLO="U")
+            layer_q.append((q_a, q_s))
+
+
+def estimate_lambda(
+    func: Callable,
+    dataloader: torch.utils.data.DataLoader,
+    eigenvectors: List[List[Tuple[torch.Tensor]]],
+    mlp_cache: List[MLPCache],
+    max_iter: int,
+) -> List[List[torch.Tensor]]:
+    lambdas = [[] for _ in range(len(mlp_cache))]
+    total_samples = 0  # record total number of samples
+    for i, batch in enumerate(dataloader):
+        # Forward pass
+        func_output = func(*batch)
+        losses, mask = (
+            func_output if isinstance(func_output, tuple) else func_output,
+            torch.tensor(1.0),
+        )
+
+        for loss in losses:
+            # Backward pass
+            loss.backward(retain_graph=True)
+
+        with torch.no_grad():
+            # Update lambdas
+            lambdas = _update_lambda(
+                lambdas,
+                mlp_cache,
+                eigenvectors,
+                total_samples,
+                mask,
+            )
+        total_samples += len(losses)
+        if i == max_iter - 1:
+            break
+
+    return lambdas
+
+
 def ifvp_at_x_ekfac(
     func: Callable,
     *x,
+    mlp_cache: Union[MLPCache, List[MLPCache]],
     in_dims: Optional[Tuple] = None,
     batch_size: int = 1,
     max_iter: Optional[int] = None,
-    mlp_cache: Union[MLPCache, List[MLPCache]],
     damping: float = 0.0,
 ) -> Callable:
     """IFVP via EK-FAC algorithm.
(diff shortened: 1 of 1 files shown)


dattri/algorithm/influence_function.py
```diff
@@ -4,6 +4,8 @@
 
 from typing import TYPE_CHECKING
 
+from dattri.task import AttributionTask
```
jiaqima, line 7: This line is likely not needed

dattri/algorithm/influence_function.py
```diff
@@ -815,3 +817,127 @@ def _get_layer_wise_reps(self,
             query_layers.append(query[:, current_idx : split_index[i] + current_idx])
             current_idx += split_index[i]
         return query_layers
+
+
+class IFAttributorEKFAC(BaseInnerProductAttributor):
+    """The inner product attributor with DataInf inverse hessian transformation."""
```
jiaqima, line 823: DataInf -> EKFAC

jiaqima:
Hi @sx-liu , thanks for the PR. Could you try to add some more detailed test about the attributor's internal functions. Could be something like this: https://github.com/TRAIS-Lab/dattri/blob/aaf192826f5435102a45f86bc8574d86aeaf34a6/test/dattri/algorithm/test_influence_function.py#L250

sx-liu:
Hi @jiaqima , I added one additional unit test for transformed test rep, which is looking at the correlation with ground truth. I think it might be hard to otherwise check the correctness about the FIM. What do you think?

dattri/algorithm/influence_function.py
```diff
@@ -815,3 +815,187 @@ def _get_layer_wise_reps(self,
...
+class IFAttributorEKFAC(BaseInnerProductAttributor):
+    """The inner product attributor with EK-FAC inverse hessian transformation."""
+
+    def __init__(self,
+                 task: AttributionTask,
+                 module_name: Optional[Union[str, List[str]]] = None,
+                 device: Optional[str] = "cpu",
+                 damping: float = 0.0,
+    ) -> None:
+        """Initialize the DataInf inverse Hessian attributor.
```
jiaqima, line 829: DataInf -> EKFAC

Hessian -> FIM

dattri/algorithm/influence_function.py
```diff
@@ -815,3 +815,187 @@ def _get_layer_wise_reps(self,
             query_layers.append(query[:, current_idx : split_index[i] + current_idx])
             current_idx += split_index[i]
         return query_layers
+
+
+class IFAttributorEKFAC(BaseInnerProductAttributor):
+    """The inner product attributor with EK-FAC inverse hessian transformation."""
```
jiaqima, line 821: hessian -> FIM

dattri/algorithm/influence_function.py
```diff
@@ -815,3 +815,187 @@ def _get_layer_wise_reps(self,
...
+        Raises:
+            ValueError: If there are multiple checkpoints in `task`.
+        """
+        super().__init__(task, None, device)
+        if len(self.task.checkpoints) > 1:
+            error_msg = ("Received more than one checkpoint. "
+                         "Ensemble of EK-FAC is not supported.")
+            raise ValueError(error_msg)
+
+        if not module_name:
```
jiaqima, line 852: if module_name is None

dattri/algorithm/influence_function.py
```diff
@@ -815,3 +815,187 @@ def _get_layer_wise_reps(self,
...
+
+        self.damping = damping
+        self.name_to_module = {
+            name: self.task.model.get_submodule(name) for name in module_name
+        }
+        self.module_to_name = {v: k for k, v in self.name_to_module.items()}
+
+        self.layer_cache = {}  # cache for each layer
+
+        def _ekfac_hook(module: torch.nn.Module,
```
jiaqima, line 874: Please add a docstring to explain the requirements for these arguments

dattri/algorithm/influence_function.py
```diff
@@ -815,3 +815,187 @@ def _get_layer_wise_reps(self,
...
+        for name in module_name:
+            mod = task.model.get_submodule(name)
+            self.handles.append(mod.register_forward_hook(_ekfac_hook))
+
+    def cache(
+        self,
+        full_train_dataloader: DataLoader,
+        max_iter: Optional[int] = None,
+    ) -> None:
+        """Cache the dataset and statistics for inverse hessian/fisher calculation.
```
jiaqima, line 900: hessian/fisher -> FIM

dattri/algorithm/influence_function.py
```diff
@@ -815,3 +815,187 @@ def _get_layer_wise_reps(self,
...
+
+        Cache the full training dataset as other attributors.
+        Estimate and cache the covariance matrices, eigenvector matrices
+        and corrected eigenvalues based on the distribution of training data.
```
jiaqima, line 904: based on the distribution of training data -> based on samples of training data.

dattri/algorithm/influence_function.py
```diff
@@ -815,3 +815,187 @@ def _get_layer_wise_reps(self,
...
+
+        Args:
+            full_train_dataloader (DataLoader): The dataloader
+                with full training samples for inverse hessian calculation.
```
jiaqima, line 908: hessian -> FIM

dattri/algorithm/influence_function.py
```diff
@@ -815,3 +815,187 @@ def _get_layer_wise_reps(self,
...
+            max_iter (int, optional): An integer indicating the maximum number of
+                batches that will be used for estimating the the covariance matrices
+                and lambdas.
```
jiaqima, line 911: default

dattri/algorithm/influence_function.py
```diff
@@ -815,3 +815,187 @@ def _get_layer_wise_reps(self,
...
+
+            if isinstance(outputs, tuple):
+                outputs = outputs[0]
+
+            outputs.retain_grad()
+            name = self.module_to_name[module]
+            # Cache the inputs and outputs
+            self.layer_cache[name] = (inputs, outputs)
+
+        self.handles = []
```
jiaqima, line 890: Move this part to self.cache()? It seems that self.handles are only needed in self.cache()

dattri/algorithm/influence_function.py
```diff
@@ -815,3 +815,187 @@ def _get_layer_wise_reps(self,
...
+        if ckpt_idx != 0:
+            error_msg = ("EK-FAC only supports single model checkpoint, "
+                         "but receives non-zero `ckpt_idx`.")
+            raise ValueError(error_msg)
+
+        # Unflatten the test_rep
+        full_model_params = {
+            k: p for k, p in self.task.model.named_parameters() if p.requires_grad
+        }
+        partial_model_params = {
```
jiaqima, line 976: How about using self.task.get_param(layer_name=self.layer_name, layer_split=True)

jiaqima (changes requested):

dattri/algorithm/influence_function.py
sx-liu: I think it's a bit hard to use this function, because it only provides the flattened parameters? Here we need the original shape information for each layer.

jiaqima: I think layer_split = True will give you a map to the module name.

Maybe @TheaperDeng has better idea here.

dattri/algorithm/influence_function.py
```diff
@@ -815,3 +815,194 @@ def _get_layer_wise_reps(self,
...
+            """Hook function for caching the inputs and outputs of a module.
+
+            Args:
+                module (torch.nn.Module): The module to which the hook is registered.
+                inputs (Union[Tensor, Tuple[Tensor]]): The module input tensor(s).
+                outputs (Union[Tensor, Tuple[Tensor]]): The module output tensor(s).
+            """
+            # Unpack tuple outputs if necessary
+            if isinstance(inputs, tuple):
+                inputs = inputs[0]
```
TheaperDeng, line 916: [bias issue] Here, we are using this `inputs` as the `a_prev` in our calculation for cov and lambda. While here we should append a `torch.ones` to the input to handle the bias.

dattri/algorithm/influence_function.py
```diff
@@ -815,3 +815,194 @@ def _get_layer_wise_reps(self,
...
+            size = math.prod(params.shape)
+            layer_test_rep[name] = test_rep[
+                :, current_index : current_index + size,
+            ].reshape(-1, *params.shape)
+            current_index += size
+
+        ifvp = {}
+
+        for name in self.module_name:
+            _v = layer_test_rep[name + ".weight"]
```
TheaperDeng, line 998: [bias issue] again, I think we need the ".bias".

dattri/func/fisher.py
```diff
@@ -405,146 +328,118 @@ def _estimate_lambda(
...
+) -> Dict[str, Tuple[torch.tensor]]:
+    """Estimate the 'covariance' matrices S and A in EK-FAC IFVP.
 
     Args:
         func (Callable): A Python function that takes one or more arguments.
             Must return the following,
-            - losses: a tensor of shape (batch_size,).
+            - loss: a single tensor of loss. Should be the mean loss by the
+                    batch size.
             - mask (optional): a tensor of shape (batch_size, t), where 1's
```
TheaperDeng, line 398: maybe we need to make this mask output clear in the init function's document of EKFAC IF Attributor

TheaperDeng (changes requested):

dattri/algorithm/influence_function.py
TheaperDeng: Since I think this can be done by easily change to the `get_param`. So I think I can handle this in next PR, and leave it as it is now in this PR.

TheaperDeng:
The EKFAC implementation overall looks good to me.

One comment is that it use `loss.backward` so we can not close the autograd during the attribution process. This is not consistent with other attribution, but I think it's fine for now.

Another comment is about the bias term, I have made some comments to the place I think need to change.

sx-liu:
I just incorporated the bias into the gradient and FIM calculation, but I found that the correlation drops drastically. I will try some other metrics and double check the implementation.

TheaperDeng (approved):
Thanks @sx-liu , the last bug was really hard to identify. LGTM

jiaqima (approved):

PR #146, opened 2024-10-17
[installation] fix installation and some name typo

TheaperDeng:
### 1. Motivation and Context

1. Change the installation option from `[all]` to `[fast_jl]` since it's the only optional package among the dependencies.
2. Rename an example filename to better reflect the content

### 2. Summary of the change

See motivation


README.md
```diff
@@ -26,20 +26,20 @@ git clone https://github.com/TRAIS-Lab/dattri
 pip install -e .
 ```
 
-If you want to use all features on CUDA and accelerate the library, you may install the full version by
+If you want to use `fast_jl` to accelerate the random projection, you may install the full version by
```
jiaqima, line 29: "the full version" -> "the version with `fast_jl`"

dattri/metric/ground_truth.py
```diff
@@ -160,7 +160,8 @@ def target_func(ckpt_path, dataloader):
             target function calculated on all test samples under `num_subsets` models,
             each retrained on a subset of the training data. The second tensor has the
             shape (num_subsets, subset_size), where each row refers to the indices of
-            the training samples used to retrain the model.
+            the training samples used to retrain the model. The targeted value will be
```
jiaqima, line 163: targeted -> target

jiaqima (approved):
Otherwise LGTM

PR #147, opened 2024-10-17
[Document] support multiple document at the same time.

TheaperDeng:
### 1. Motivation and Context
We will have multiple document checkpoints for each released version as well as the latest version.

### 2. Summary of the change
Add a new step in `doc.yml` called Deploy-not-main. This will help us to have
1. The latest document (reflect the API of `main` branch) still show on `https://trais-lab.github.io/dattri`
2. The released version document will be shown on `https://trais-lab.github.io/dattri/<branch-name>`


xingjian-zhang (approved):
LGTM.

Issue #154, opened 2024-11-29
[dattri.trak] Improving memory efficiency when calculating TRAK


KurisuTheAmadeus:
Current Implementation of TRAK calculates final result using
```
    return (running_xinv_XTX_XT @ self.Q.diag().to(self.device)).T
```
and the introduction of diag matrix creates large memory overhead, especially in the case when number of training data points is large.
A possible fix is as follows
```
        if train_dataloader is not None:
            return (running_xinv_XTX_XT * running_Q.to(self.device).unsqueeze(0)).T
        return (running_xinv_XTX_XT * self.Q.to(self.device).unsqueeze(0)).T
```
and I have verified the equivalency using two different setups.

tingwl0122:
make sense to me

TheaperDeng:
Could you directly post a PR for this? The code looks good to me.

KurisuTheAmadeus:
I opened a PR regarding this, see #155. Thanks!

PR #155, opened 2024-12-06
feat: more memory-efficient TRAK calculation

KurisuTheAmadeus:
This is a pull request corresponding to issue #154. Basically it replaces the calculation of diag(Q) in TRAK with a more memory-efficient one. The code is exactly the same as I proposed  in #154


TheaperDeng (approved):
LGTM

PR #157, opened 2024-12-13
[dattri] Add huggingface example (GPT-2 + WikiText-2) and related bug fix

TheaperDeng:
### 1. Motivation and Context

- Models from huggingface `transformers` is widely used and some additional fix should be included in `dattri`.
- A larger experiment is added (GPT-2 + WikiText-2) to `experiments/`

### 2. Summary of the change

Experiment added:
- An example GPT-2 + WikiText-2 is added to `experiments/gpt2_wikitext`. Please check https://github.com/TheaperDeng/dattri-public/tree/gpt2_wiki2_hf/experiments/gpt2_wikitext for more detail.

Example added
- An example using Dropout Ensemble in https://github.com/TheaperDeng/dattri-public/blob/gpt2_wiki2_hf/examples/pretrained_benchmark/trak_dropout_lds.py

API change (backlog supported, no need for user code change):
- Add `checkpoints_load_func` to `AttributionTask`, The parameter is used for some model that has special loading strategies, e.g., huggingface model. A typical example for huggingface model is

```python
def checkpoints_load_func(model, checkpoint):
    model = AutoModelForCausalLM.from_pretrained(checkpoint).cuda()
    model.eval()
    return model

task = AttributionTask(loss_func=f, model=model,
                       checkpoints=checkpoints,
                       checkpoints_load_func=checkpoints_load_func)  # if not specific, a default one will be applied.
```

Bug fix:
- For `TRAKAttributor`
   - The scaling number calculation is changed to a much more efficient implementation.
   - The "pre-gradient" calculation is changed to support non-tuple/list data format (e.g., dict, that is widely used for models in `transformers`
   - Change batch size calculation method to the shape[0] of gradient rather than input data

TODO:
- Better "pre-gradient" calculation implementation for TRAK and other attributors. (has been recorded to source code).

### DIFF
--- a/.github/workflows/examples_test.yml
+++ b/.github/workflows/examples_test.yml
@@ -27,6 +27,8 @@ jobs:
         python examples/noisy_label_detection/trak_noisy_label.py --device cpu
         python examples/pretrained_benchmark/influence_function_lds.py --device cpu
         python examples/pretrained_benchmark/trak_loo.py --device cpu
+        sed -i 's/* 10/* 2/g' examples/pretrained_benchmark/trak_dropout_lds.py
+        python examples/pretrained_benchmark/trak_dropout_lds.py --device cpu
         python examples/brittleness/mnist_lr_brittleness.py --method cg --device cpu
     - name: Uninstall the package
       run: |
--- a/dattri/algorithm/trak.py
+++ b/dattri/algorithm/trak.py
@@ -72,7 +72,12 @@ class TRAKAttributor(BaseAttributor):
         """
         self.task = task
         self.norm_scaler = (
-            sum(p.numel() for p in self.task.get_param(ckpt_idx=0)[0]) ** 0.5
+            sum(
+                p.numel()
+                for _, p in self.task.get_model().named_parameters()
+                if p.requires_grad
+            )
+            ** 0.5
         )
         self.projector_kwargs = DEFAULT_PROJECTOR_KWARGS
         if projector_kwargs is not None:
@@ -84,6 +89,7 @@ class TRAKAttributor(BaseAttributor):
         self.correct_probability_func = vmap(
             correct_probability_func,
             in_dims=(None, 0),
+            randomness="different",
         )
         self.full_train_dataloader = None
 
@@ -127,14 +133,22 @@ class TRAKAttributor(BaseAttributor):
                 desc="calculating gradient of training set...",
                 leave=False,
             ):
-                train_batch_data = tuple(data.to(self.device) for data in train_data)
+                # TODO: reorganize the data pre-grad processing.
+                if isinstance(train_data, (tuple, list)):
+                    train_batch_data = tuple(
+                        data.to(self.device) for data in train_data
+                    )
+                else:
+                    train_batch_data = train_data
+
                 grad_t = self.grad_loss_func(parameters, train_batch_data)
                 grad_t = torch.nan_to_num(grad_t)
                 grad_t /= self.norm_scaler
+                batch_size = grad_t.shape[0]
                 grad_p = (
                     random_project(
                         grad_t,
-                        train_batch_data[0].shape[0],
+                        batch_size,
                         **self.projector_kwargs,
                     )(grad_t, ensemble_id=ckpt_idx)
                     .clone()
@@ -143,7 +157,7 @@ class TRAKAttributor(BaseAttributor):
                 full_train_projected_grad.append(grad_p)
                 Q.append(
                     (
-                        torch.ones(train_batch_data[0].shape[0]).to(self.device)
+                        torch.ones(batch_size).to(self.device)
                         - self.correct_probability_func(
                             _unflatten_params(full_parameters, self.task.get_model()),
                             train_batch_data,
@@ -167,7 +181,7 @@ class TRAKAttributor(BaseAttributor):
         self.inv_XTX_XT_list = inv_XTX_XT_list
         self.Q = running_Q
 
-    def attribute(  # noqa: PLR0912
+    def attribute(  # noqa: PLR0912,PLR0915
         self,
         test_dataloader: torch.utils.data.DataLoader,
         train_dataloader: Optional[torch.utils.data.DataLoader] = None,
@@ -239,20 +253,26 @@ class TRAKAttributor(BaseAttributor):
                     desc="calculating gradient of training set...",
                     leave=False,
                 ):
-                    train_batch_data = tuple(
-                        data.to(self.device) for data in train_data
-                    )
+                    # TODO: reorganize the data pre-grad processing.
+                    if isinstance(train_data, (tuple, list)):
+                        train_batch_data = tuple(
+                            data.to(self.device) for data in train_data
+                        )
+                    else:
+                        train_batch_data = train_data
+
                     grad_t = self.grad_loss_func(
                         parameters,
                         train_batch_data,
                     )
                     grad_t = torch.nan_to_num(grad_t)
                     grad_t /= self.norm_scaler
+                    batch_size = grad_t.shape[0]
 
                     grad_p = (
                         random_project(
                             grad_t,
-                            train_batch_data[0].shape[0],
+                            batch_size,
                             **self.projector_kwargs,
                         )(grad_t, ensemble_id=ckpt_idx)
                         .clone()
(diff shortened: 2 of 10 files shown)


dattri/algorithm/trak.py
```diff
@@ -84,6 +89,7 @@ def m(params, image_label_pair):
         self.correct_probability_func = vmap(
             correct_probability_func,
             in_dims=(None, 0),
+            randomness="different",
```
tingwl0122, line 92: I am wondering why we suddenly need this? I see there are also other places adding this. Does this affect the performance much?

dattri/algorithm/trak.py
```diff
@@ -127,14 +133,22 @@ def cache(
                 desc="calculating gradient of training set...",
                 leave=False,
             ):
-                train_batch_data = tuple(data.to(self.device) for data in train_data)
+                # TODO: reorganize the data pre-grad processing.
+                if isinstance(train_data, (tuple, list)):
+                    train_batch_data = tuple(
+                        data.to(self.device) for data in train_data
+                    )
+                else:
```
tingwl0122, line 141: per your description, so if `train_data` is not a tuple of a list, then it must be a dict?
And why if `train_data` is a dict, then we don't need to put data onto `self.device`?
In addition, do we need to restrict the type of `train_data` here?

dattri/algorithm/trak.py
```diff
@@ -239,20 +253,26 @@ def attribute(  # noqa: PLR0912
...
-                        data.to(self.device) for data in train_data
-                    )
+                    # TODO: reorganize the data pre-grad processing.
+                    if isinstance(train_data, (tuple, list)):
+                        train_batch_data = tuple(
+                            data.to(self.device) for data in train_data
+                        )
+                    else:
+                        train_batch_data = train_data
+
```
tingwl0122, line 263: the same issue as above

dattri/algorithm/trak.py
```diff
@@ -282,15 +302,19 @@ def attribute(  # noqa: PLR0912
                 desc="calculating gradient of test set...",
                 leave=False,
             ):
-                test_batch_data = tuple(data.to(self.device) for data in test_data)
+                # TODO: reorganize the data pre-grad processing.
+                if isinstance(test_data, (tuple, list)):
+                    test_batch_data = tuple(data.to(self.device) for data in test_data)
+                else:
+                    test_batch_data = test_data
```
tingwl0122, line 309: the same issue as above

dattri/task.py
```diff
@@ -75,8 +95,19 @@ def f(params, data):
                     loss = nn.CrossEntropyLoss()
                     yhat = torch.func.functional_call(model, params, image)
                     return loss(yhat, label)
+                ```
+            checkpoints_load_func (Callable): The checkpoint load function.
+                The input is optional, if not provided, the checkpoint load
+                function will be a default one using model.load_state_dict.
+                The parameter is used for some model that has special
```
tingwl0122, line 102: some models that have special...

dattri/task.py
```diff
@@ -90,6 +121,11 @@ def f(params, data):
         self.original_target_func = target_func
         self.target_func = flatten_func(self.model)(target_func)
 
+        if checkpoints_load_func is None:
+            self.checkpoints_load_func = _default_checkpoint_load_func
+        else:
+            self.checkpoints_load_func = checkpoints_load_func
+
```
tingwl0122, line 128: do we need to be careful here? now it seems like we are either using default or huggingface loading.
Will there be any other type of loading?

dattri/task.py
```diff
@@ -123,20 +167,14 @@ def _load_checkpoints(self, ckpt_idx: int) -> None:
...
-                self.model.load_state_dict(self.checkpoints[ckpt_idx])
+            self.model = self.checkpoints_load_func(
+                self.model,
+                self.checkpoints[ckpt_idx],
+            )
             self.current_checkpoint_idx = ckpt_idx
             self.named_parameters = {
                 k: p for k, p in self.model.named_parameters() if p.requires_grad
             }
-        self.model.eval()
```
tingwl0122, line 139: what is the original purpose of this line?

examples/pretrained_benchmark/trak_dropout_lds.py
```diff
@@ -0,0 +1,87 @@
...
+        score = attributor.attribute(
+            DataLoader(
+                model_details["test_dataset"],
+                batch_size=5000,
+                sampler=model_details["test_sampler"],
+            ),
+        )
+
+    lds_score = lds(score, groundtruth)[0]
+    print("lds:", torch.mean(lds_score[~torch.isnan(lds_score)]))
```
tingwl0122, line 87: this example looks to me if we have validated its performance gain matches our expectation.

tingwl0122 (commented):
I read through the files other than the `experiments/gpt2_wikitext` folder.
My only concern is that whether we need to unify this experiment into existing ones.
Does the original nanogpt resembles gpt2 used here?
If yes, then I guess it is just nanogpt + (ShakeSpeare or Wikitext)?

.github/workflows/examples_test.yml
```diff
@@ -27,6 +27,8 @@ jobs:
         python examples/noisy_label_detection/trak_noisy_label.py --device cpu
         python examples/pretrained_benchmark/influence_function_lds.py --device cpu
         python examples/pretrained_benchmark/trak_loo.py --device cpu
+        sed -i 's/* 10/* 2/g' examples/pretrained_benchmark/trak_dropout_lds.py
```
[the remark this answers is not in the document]
TheaperDeng: This line is using `sed` to change the file (only during the test) and reduce the computational cost (ensemble number changed from 10 to 2). It will not affect the raw file.

dattri/algorithm/trak.py
TheaperDeng: No, it does not affect the performance. But if we use dropout enabled during the computation, we need to state this `randomness="different“` or there will be error (defaultly, `randomness="error"`). Another choice is `randomness="same"`, this will affect the performance of dropout ensemble.

dattri/algorithm/trak.py
TheaperDeng: I leave this to future PR. As I stated in TODO (`TODO: reorganize the data pre-grad processing.`). I will raise a PR for future developers to do this. I understand now the preprocessing is very weak but it seems too much for this PR.

dattri/algorithm/trak.py
TheaperDeng: same

dattri/algorithm/trak.py
TheaperDeng: same

dattri/task.py
TheaperDeng: The logic of this code block is. If `checkpoints_load_func` is not explicitly provided, we will use `_default_checkpoint_load_func`. `checkpoints_load_func` is provided by users, it can be huggingface loading or any other things.

dattri/task.py
TheaperDeng: When model is loaded, it will be changed to eval mode. Now it's in `_default_checkpoint_load_func`.

examples/pretrained_benchmark/trak_dropout_lds.py
TheaperDeng: Yes, it has been validated.

.github/workflows/examples_test.yml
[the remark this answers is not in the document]
tingwl0122: got it

dattri/algorithm/trak.py
tingwl0122: Got it. Just curious, so if we state `randomness="same"`, there will only be one dropout version basically?

dattri/algorithm/trak.py
tingwl0122: Make sense

dattri/task.py
tingwl0122: Oh got it. sorry for the misunderstanding

dattri/task.py
tingwl0122: got it

tingwl0122 (approved):
LGTM now
so should we raise an issue about the gpt2+wiki setting reorganization?

dattri/algorithm/trak.py
TheaperDeng: I think it will be one dropout version for each independent checkpoint. So there is no meaning to use dropout ensemble anymore.

TheaperDeng:
> LGTM now so should we raise an issue about the gpt2+wiki setting reorganization?

Yes, I will raise it after merging the PR.

PR #158, opened 2024-12-13
[dattri/examples] add data cleaning example

TheaperDeng:
### 1. Motivation and Context

Add an example for data selection.

### 2. Summary of the change

Add an example for data selection. This script find the top-K low-quality data that "harm" the validation performance the most. The script also evaluate the test performance after removing the top-K low-quality data points.

```bash
Namespace(method='explicit', device='cuda', seed=0, train_size=10000, val_size=10000, test_size=5000, remove_number=100)
Test loss: 0.39373043179512024                                                                                                                                                        
Accuracy: 88.52                                                                                                                                                                       
Test loss (after): 0.38540247082710266
Accuracy  (after): 88.94
```

### DIFF
--- a/examples/data_selection/influence_function_data_selection.py
+++ b/examples/data_selection/influence_function_data_selection.py
@@ -1,4 +1,4 @@
-"""This example shows how to use the IF to detect noisy labels in the MNIST."""
+"""This example shows how to use the IF to detect low-quality data in the MNIST."""
 
 import argparse
 from functools import partial
--- a/examples/data_selection/influence_function_data_selection.py
+++ b/examples/data_selection/influence_function_data_selection.py
@@ -1,0 +1,151 @@
+"""This example shows how to use the IF to detect noisy labels in the MNIST."""
+
+import argparse
+from functools import partial
+
+import torch
+from torch import nn
+
+from dattri.algorithm.influence_function import (
+    IFAttributorArnoldi,
+    IFAttributorCG,
+    IFAttributorDataInf,
+    IFAttributorExplicit,
+    IFAttributorLiSSA,
+)
+from dattri.benchmark.datasets.mnist import create_mnist_dataset, train_mnist_lr
+from dattri.benchmark.utils import SubsetSampler
+from dattri.metric import mislabel_detection_auc
+from dattri.task import AttributionTask
+
+ATTRIBUTOR_MAP = {
+    "explicit": partial(IFAttributorExplicit, regularization=0.01),
+    "cg": partial(IFAttributorCG, regularization=0.01),
+    "lissa": partial(IFAttributorLiSSA, recursion_depth=100),
+    "datainf": partial(IFAttributorDataInf, regularization=0.01),
+    "arnoldi": partial(IFAttributorArnoldi, regularization=0.01, max_iter=10),
+}
+
+
+if __name__ == "__main__":
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--method", type=str, default="explicit")
+    parser.add_argument("--device", type=str, default="cuda")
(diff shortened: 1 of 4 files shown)


examples/readme.md
```diff
@@ -23,3 +23,9 @@ This section contains examples using the pretrained checkpoints and pre-calculat
 This section contains examples using attribution scores to estimate the brittleness of a model.
 
 [Use influence function to estimate the brittleness of losigitc regression trained on Mnist10](./brittleness/mnist_lr_brittleness.py)
+
+## Data selection
```
tingwl0122, line 27: I guess we can call this application "data cleaning"?
which is more consistent to what we understand (data selection if more like the opposite of this) and also our survey paper's tone

examples/data_selection/influence_function_data_selection.py
```diff
@@ -0,0 +1,151 @@
...
+        sampler=SubsetSampler(range(args.train_size)),
+    )
+    val_loader = torch.utils.data.DataLoader(
+        dataset,
+        batch_size=1000,
+        sampler=SubsetSampler(range(args.train_size, args.train_size + args.val_size)),
+    )
+    test_loader = torch.utils.data.DataLoader(
+        dataset_test,
+        batch_size=1,
```
tingwl0122, line 67: why we suddenly use such a small batch size?

examples/data_selection/influence_function_data_selection.py
```diff
@@ -0,0 +1,151 @@
...
+    value, indices = torch.sort(score_overall, descending=True)
+
+    # calculate the test loss
+    loss = nn.CrossEntropyLoss()
+    loss_value = []
+    for image, label in test_loader:
+        image, label = image.to(args.device), label.to(args.device)
+        yhat = model(image)
+        loss_value.append(loss(yhat, label.long()).item())
+    loss_value = torch.mean(torch.tensor(loss_value))
```
tingwl0122, line 102: Now I understand why we are using `batch_size = 1` for test loader.
I guess we can refer to https://github.com/TRAIS-Lab/dattri/blob/a73864f606df98a9b6f8abfbf7b7418b53c9707d/dattri/benchmark/datasets/mnist/mnist_mlp.py#L53 to allow normal batch size?

examples/data_selection/influence_function_data_selection.py
```diff
@@ -0,0 +1,151 @@
...
+    model.eval()
+
+    # calculate the test loss
+    loss = nn.CrossEntropyLoss()
+    loss_value = []
+    for image, label in test_loader:
+        image, label = image.to(args.device), label.to(args.device)
+        yhat = model(image)
+        loss_value.append(loss(yhat, label.long()).item())
+    loss_value = torch.mean(torch.tensor(loss_value))
```
tingwl0122, line 137: same as above

examples/readme.md
TheaperDeng: good idea. I can change it to data cleaning for better understanding of our users.

examples/data_selection/influence_function_data_selection.py
TheaperDeng: well I guess I can make this larger. It does not affect the TDA runtime since only validation dataset are used for TDA.

examples/readme.md
tingwl0122: solved

examples/data_selection/influence_function_data_selection.py
tingwl0122: make sense

tingwl0122 (commented):
LGTM

tingwl0122 (approved):
LGTM

Issue #164, opened 2025-01-15
implementation of RelatIF


TheaperDeng:
implement two versions of relatIF in https://arxiv.org/pdf/2003.11630.

No need to make them new attributors. I think a new method/config parameter on current IF attributors would be enough.

Also, some examples need to be added.

jxbb824:
For this task, we need to compute the training point’s IHVP or self-influence, which can be done via the existing `transform_test_rep` method.

In terms of implementation, I plan to add an `attribute_relatif` method to `BaseInnerProductAttributor`. By default, it will return the same result as `attribute`. Then, in each specific influence-function subclass, I will implement its detailed behavior. 

It might be possible to implement `attribute_relatif` entirely within `BaseInnerProductAttributor` (similar to how `attribute` is handled) so that all subclasses share the same logic. However, for now, I will avoid that approach to reduce the risk of introducing additional bugs. I will first implement the `explicit`, `cg`, and `lissa` versions.

TheaperDeng:
Thank you @jxbb824 . Could you try to organize the design by answering the following questions?

1. what methods (external methods, i.e., the ones that will be directly called by users) will I add/modify. Use docstring directly to answer the question.

Example answer
```python
...
def function_new(self, a, b):
'''This function is used to calculate two numbers' summation.

Args:
    a (int): the first number
    b (int): the second number

Returns:

    (int): the summation of a and b
...
```

2. How will users use this function? Please use a very simple (but comprehensive) example call to answer the question
Example answer
```python
calculator = Calculator()

c = calculator.function_new(a,b)
```

3. How will I implement it?

jxbb824:
1. what methods (external methods, i.e., the ones that will be directly called by users) will I add/modify. Use docstring directly to answer the question.
   
   ```python
   def attribute(
       self,
       train_dataloader: DataLoader,
       test_dataloader: DataLoader,
       relativeIF_method: Optional[str] = None,
   ) -> torch.Tensor:
       """Calculate the influence of the training set on the test set.
   
       Args:
           ......
           relativeIF_method (Optional[str]): Method for normalizing the influence values. 
               Supported options:
               - `"l"`: Normalizes by `sqrt(g_i^T (H^-1 g_i))`.
               - `"theta"`: Normalizes by `||H^-1 g_i||`.
               - `None`: No normalization applied.
   
       Returns:
           torch.Tensor: The influence of the training set on the test set, with
               the shape of (num_train_samples, num_test_samples).
[...]

PR #170, opened 2025-02-05
Implement RelatIF for Explicit, CG, and LISSA

jxbb824:
### 1. Motivation and Context  
This PR implements two versions of RelatIF as described in [https://arxiv.org/pdf/2003.11630](https://arxiv.org/pdf/2003.11630). Instead of creating new attributors, the implementation integrates RelatIF into the existing influence function attributors via a new config parameter.  

This addresses issue **#164**.

### 2. Summary of the change  
- Implemented two versions of RelatIF for the **Explicit**, **CG**, and **LISSA** influence function attributors.  
- Introduced a new config parameter to enable RelatIF computation within the existing framework.  
- Added an example that replicates the first figure from the RelatIF paper to demonstrate the correctness of the implementation.  

- [x] Unit test: Added tests to verify RelatIF computation in explicit, CG, and LISSA attributors.  
- [x] Application test: Included an example script that reproduces the first figure from the RelatIF paper.  
- [ ] Document test: Verified that the new method/config parameter is correctly documented.

### DIFF
--- a/dattri/algorithm/base.py
+++ b/dattri/algorithm/base.py
@@ -259,6 +259,7 @@ class BaseInnerProductAttributor(BaseAttributor):
         self,
         train_dataloader: DataLoader,
         test_dataloader: DataLoader,
+        relativeif_method: Optional[str] = None,
     ) -> torch.Tensor:
         """Calculate the influence of the training set on the test set.
 
@@ -270,6 +271,12 @@ class BaseInnerProductAttributor(BaseAttributor):
                 not be shuffled.
             test_dataloader (DataLoader): Dataloader for test samples to calculate
                 the influence. The dataloader should not be shuffled.
+            relativeif_method (Optional[str]): Method for normalizing the
+                influence values.
+                Supported options:
+                - `"l"`: Normalizes by `sqrt(g_i^T (H^-1 g_i))`.
+                - `"theta"`: Normalizes by `||H^-1 g_i||`.
+                - `None`: No normalization applied.
 
         Returns:
             torch.Tensor: The influence of the training set on the test set, with
@@ -314,6 +321,15 @@ class BaseInnerProductAttributor(BaseAttributor):
                     ckpt_idx=checkpoint_idx,
                     data=train_batch_data,
                 )
+
+                denom = None
+                if relativeif_method is not None:
+                    denom = self._compute_denom(
+                        checkpoint_idx,
+                        train_batch_rep,
+                        relativeif_method=relativeif_method,
+                    )
+
                 # transform the train representations
                 train_batch_rep = self.transform_train_rep(
                     ckpt_idx=checkpoint_idx,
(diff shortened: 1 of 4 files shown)


dattri/algorithm/influence_function.py
```diff
@@ -289,6 +289,49 @@ def transform_test_rep(
...
+                - `None`: Raises an error.
+
+        Returns:
+            torch.Tensor: The computed denominator for normalization. It is a
+            1-d dimensional tensor with the shape of (batch_size).
+
+        Raises:
+            ValueError: If an unknown `method` is provided.
+        """
+        transformed = self.transform_test_rep(ckpt_idx, train_batch_rep)
```
TheaperDeng, line 317: I don't know if this `transformed` can be used for future computation just to save the computational resource. (I mean it definitely can, but we may have a messy code base after that) **But I think it's fine to put it as it is right now**.

dattri/algorithm/base.py
```diff
@@ -314,6 +321,15 @@ def attribute(
                     ckpt_idx=checkpoint_idx,
                     data=train_batch_data,
                 )
+
+                denom = None
+                if relativeif_method is not None:
+                    denom = self._compute_denom(
+                        checkpoint_idx,
+                        train_batch_rep,
```
TheaperDeng, line 329: I think you need 2 input here. One is this `train_batch_rep` comes from `train_batch_rep = self.generate_train_rep(..., train_batch_data, ...)` and another `test_batch_rep` comes from `self.generate_test_rep(..., train_batch_data, ...)`. The first one will be used for H^-1g_i and the second one will be used as g_i when `relativeif_method == "l"`. Since sometimes there is a difference between loss function and target function.

TheaperDeng (commented):
good job. Please change according to the comments. Another problem is that how much overhead does it involve when open the `relatif_method` config to `l` and `theta`.

TheaperDeng:
Oh another minor comment is that you may add RelatIF here: https://github.com/TheaperDeng/dattri?tab=readme-ov-file#algorithms-supported and remove the TODO here: https://github.com/TRAIS-Lab/dattri?tab=readme-ov-file#development-plan

TheaperDeng:
Since there is an example added to the repo, please add the example test script here:
https://github.com/TRAIS-Lab/dattri/blob/dfdf621330e694c96e8eb085ac4ba462372db1da/.github/workflows/examples_test.yml#L22-L33

dattri/algorithm/influence_function.py
jxbb824: Thanks for the feedback! I'll keep it as is for now.

dattri/algorithm/base.py
jxbb824: Thanks for the explanation! Since the implementation in the Relatif paper doesn’t use `s_test`, it seems that the roles of `transform_test_rep` and `transform_train_rep` are effectively reversed compared to how other influence functions are implemented in the library. While this doesn’t cause any issues for now, it could introduce potential problems if we later extend this to other methods where `transform_train_rep` has an actual purpose. It might also lead to some confusion in understanding the code.

Additionally, the naming of `self.generate_test_rep(..., train_batch_data, ...)` can be a bit misleading. For now, I’m just calling it `test_batch_rep`.

jxbb824:
In my example, I used `matplotlib` for plotting. Will this cause any issues during the testing process?

TheaperDeng:
> In my example, I used `matplotlib` for plotting. Will this cause any issues during the testing process?

Please add any dependencies you need for testing here: https://github.com/TRAIS-Lab/dattri/blob/02aa0906388d5dec420e7366cfffd7ab77a94e69/pyproject.toml#L32

dattri/algorithm/base.py
```diff
@@ -314,6 +321,20 @@ def attribute(
                     ckpt_idx=checkpoint_idx,
                     data=train_batch_data,
                 )
+
+                denom = None
+                if relatif_method is not None:
+                    test_batch_rep = self.generate_test_rep(
+                        ckpt_idx=checkpoint_idx,
+                        data=train_batch_data,
+                    )
```
TheaperDeng, line 330: maybe make this calculation conditional on if the `relatif_method=='l'`?

TheaperDeng (commented):
others LGTM

jxbb824:
I’ve made `test_batch_rep` optional and adjusted its calculation to be conditional. I’ve also added `matplotlib` as a dependency in `pyproject.toml`, but I’m not sure if the example will pass testing since I couldn’t run it locally.

Issue #174, opened 2025-02-06
Replace all `torch.linalg.inv` by `torch.linalg.solve` in TRAK attributor


TheaperDeng:
https://github.com/TRAIS-Lab/dattri/blob/dfdf621330e694c96e8eb085ac4ba462372db1da/dattri/algorithm/trak.py#L172-L174

https://github.com/TRAIS-Lab/dattri/blob/dfdf621330e694c96e8eb085ac4ba462372db1da/dattri/algorithm/trak.py#L330-L331

`torch.linalg.solve` computes X = A.inverse() @ B in a faster and more numerically stable way than performing the computations separately.

JhaMohan:
@TheaperDeng created i pull request [175](https://github.com/TRAIS-Lab/dattri/pull/175)

jackjin1997:
Hi~ here the [pull request.](https://github.com/TRAIS-Lab/dattri/pull/177)

jiaqima:
Closed by #186

PR #175, opened 2025-02-06
Replaced all torch.linalg.inv by torch.linalg.solve

JhaMohan:
## Description
Issue #174 
Replace all torch.linalg.inv by torch.linalg.solve in TRAK attributor
As torch.linalg.solve computes X = A.inverse() @ B in a faster and more numerically stable way than performing the computations separately.


jiaqima:
Thank you for your contribution. But this is not how linalg.solve works. Closing this PR for now.

PR #177, opened 2025-02-07
perf: Replace all torch.linalg.inv by torch.linalg.solve in TRAK attr…

jackjin1997:
…ibutor #174

- [ ] ...


jiaqima:
Thank you for your contribution. But this is not how linalg.solve works. Closing this PR for now.

PR #181, opened 2025-03-15
Fix version problem of tj-actions in darglint.yml

jiaqima:
## Description

This is a small fix of tj-actions in darglint.yml. The existing version v44 seems to start failing.

### DIFF
--- a/.github/workflows/darglint.yml
+++ b/.github/workflows/darglint.yml
@@ -20,7 +20,7 @@ jobs:
         pip install darglint
     - name: Get changed files
       id: changed-files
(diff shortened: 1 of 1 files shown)


jiaqima:
It appears that tj-actions/changed-files is temporally taken down due to a security issue: https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised

Will wait and see if this will be solved soon.

jiaqima:
tj-actions/changed-files seems to have recovered. This PR is no longer needed.

PR #187, opened 2025-05-05
[hessian] Solve for issue173. Avoid torch.eye to reduce memory in ihvp_explicit in dattri/func/hessian.py

leaf-source:
## Description

Avoid torch.eye by adding regularization directly to Hessian diagonal

### DIFF
--- a/dattri/func/hessian.py
+++ b/dattri/func/hessian.py
@@ -264,8 +264,7 @@ def ihvp_explicit(
         """
         hessian_tensor = hessian_func(*x)
         return torch.linalg.solve(
-            hessian_tensor
-            + torch.eye(hessian_tensor.shape[0]).to(v.device) * regularization,
+            hessian_tensor.diagonal().add_(regularization),
             v.T,
         ).T
 
--- a/dattri/func/hessian.py
+++ b/dattri/func/hessian.py
@@ -264,7 +264,7 @@ def ihvp_explicit(
         """
         hessian_tensor = hessian_func(*x)
         return torch.linalg.solve(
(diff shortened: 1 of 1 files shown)


dattri/func/hessian.py
```diff
@@ -265,7 +265,12 @@ def _ihvp_explicit_func(x: Tuple[torch.Tensor, ...], v: Tensor) -> Tensor:
         hessian_tensor = hessian_func(*x)
         return torch.linalg.solve(
             hessian_tensor
-            + torch.eye(hessian_tensor.shape[0]).to(v.device) * regularization,
+            + torch.diag(
+                torch.full(
+                    (hessian_tensor.shape[0],),
+                    regularization,
+                    device=v.device,
+                )),
```
TheaperDeng, line 273: I am not sure, but I think `torch.diag` will create a full matrix as the shape of hessian matrix as well right?

jiaqima: Get some test results for the memory usage?

leaf-source: Sorry I think I made a mistake. It still creates a full matrix. I'll revise my code.

TheaperDeng: Yeah, @leaf-source could you also design a unit test for this case examine the memory usage as well?

leaf-source: Sure, I will.

TheaperDeng:
After careful discussion, this won't reduce the peak memory usage of the whole process since the bottleneck is somewhere else. I will close this PR as well as the issue.

PR #188, opened 2025-05-31
Issue178 solve:  creates only one iterator object instead of creating one every loop

leaf-source:
## Description

Fix improper data loader iteration in IFAttributorDataInf and IFAttributorArnoldi

Previously, both classes repeatedly called next(iter(full_train_dataloader)) inside a loop, which resulted in re-creating the iterator at each iteration and thus always sampling the first batch. This led to redundant and non-representative data being cached or processed.

This patch fixes the issue by instantiating the iterator once before the loop.


TheaperDeng (approved):
LGTM

Issue #191, opened 2025-07-28
Implementation of Data Value Embedding (DVEmb)


BilyHurington:
Implement DVEmb in https://arxiv.org/pdf/2412.09538.

1. what methods (external methods, i.e., the ones that will be directly called by users) will I add/modify. Use docstring directly to answer the question.

I plan to implement a new attributor class DVEmbAttributor, including the following methods.

```py
def __init__(
        self,
        model: nn.Module,
        loss_func: Callable,
        device: str = "cpu",
        proj_dim: Optional[int] = None,
        factorization_type: str = "none",
    ) -> None:
        """Initializes the DVEmb attributor.

        Args:
            model: The PyTorch model to be attributed.
            loss_func: A per-sample loss function. It should take model parameters
                       and a single data sample (unbatched) as input.
            device: The device to run computations on (e.g., "cpu", "cuda").
            proj_dim: The dimension for projection (if used).
            factorization_type: Type of gradient factorization to use. Options are
                                "none"(default),
                                "kronecker"(same with paper),
                                or "elementwise"(more memory-efficient).
        """
```

```py
def cache_gradients(
        self,
        epoch: int,
        batch_data: tuple[Tensor, ...],
        indices: Tensor,
        learning_rate: float,
    ) -> None:
        """Cache per-sample gradients for a specific epoch and training step.

        Args:
            epoch: The current epoch number.
            batch_data: A tuple containing the batch of inputs and targets,
                        e.g., (inputs, labels).
            indices: A tensor containing the original indices for the samples
                     in batch_data.
            learning_rate: The learning rate for this step.
        """
```

```py
def compute_embeddings( 
        self,
        gradients: Optional[Dict[int, List[Tensor]]] = None,
        learning_rates: Optional[Dict[int, List[float]]] = None,
    ) -> None:
        """Computes data value embeddings for each epoch separately.

        Args:
            gradients: Optional external gradients instead of cached ones
                (e.g., (epoch -> list of per-sample gradients)).
            learning_rates: Optional external learning rates instead of cached ones
                (e.g., (epoch -> list of learning rates)).
        """
```

```py
def attribute(
        self,
        test_dataloader: DataLoader,
        epoch: Optional[int] = None,
        traindata_indices: Optional[List[int]] = None,
    ) -> torch.Tensor:
        """Calculates influence scores for a test set for one or all epochs.

        Args:
            test_dataloader: A dataloader for the test set.
[...]

PR #193, opened 2025-08-18
DVEmbAttributor

BilyHurington:
### 1. Motivation and Context

This pull request introduces the Data Value Embedding (DVEmb) attributor, a method for calculating trajectory-specific data influence. Unlike existing methods that often overlook the temporal dynamics of model training, DVEmb captures how the influence of a data point evolves over time by creating epoch-specific embeddings, which allows for a more accurate analysis of data value.

### 2. Summary of the change

[github issue](https://github.com/TRAIS-Lab/dattri/issues/191)

### DIFF
--- a/dattri/algorithm/__init__.py
+++ b/dattri/algorithm/__init__.py
@@ -2,6 +2,7 @@
 
 from .block_projected_if.block_projected_if import BlockProjectedIFAttributor
 from .data_shapley import KNNShapleyAttributor
+from .dvemb import DVEmbAttributor
 from .factgrass import FactGraSSAttributor
 from .influence_function import (
     IFAttributorArnoldi,
@@ -15,10 +16,10 @@ from .logra import LoGraAttributor
 from .rps import RPSAttributor
 from .tracin import TracInAttributor
 from .trak import TRAKAttributor
-from .dvemb import DVEmbAttributor
 
 __all__ = [
     "BlockProjectedIFAttributor",
+    "DVEmbAttributor",
     "FactGraSSAttributor",
     "IFAttributorArnoldi",
     "IFAttributorCG",
@@ -31,5 +32,4 @@ __all__ = [
     "RPSAttributor",
     "TRAKAttributor",
     "TracInAttributor",
-    "DVEmbAttributor",
 ]
--- a/test/dattri/algorithm/test_dvemb.py
+++ b/test/dattri/algorithm/test_dvemb.py
@@ -169,11 +169,6 @@ class TestDVEmbAttributor:
         expected_proj_dim = int(math.sqrt(proj_dim / num_layers))
         assert attributor.projection_dim == expected_proj_dim
 
-        assert (
-            attributor.cached_factors[0][0][0]["A"].shape[1]
-            == attributor.projection_dim
-        )
-
     def test_dvemb_elementwise_with_projection(self):
         """Test DVEmb with elementwise factorization and projection."""
         proj_dim = 16


dattri/algorithm/dvemb.py
```diff
@@ -0,0 +1,237 @@
...
+        self.data_indices: Dict[int, List[Tensor]] = {}
+        self.embeddings: Dict[int, Tensor] = {}
+
+    def cache_gradients(
+        self,
+        epoch: int,
+        batch_data: tuple[Tensor, ...],
+        indices: Tensor,
+        learning_rate: float,
+    ) -> None:
```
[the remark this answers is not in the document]
BilyHurington: It should be called at every training step, not just at the end of the epoch, self.cached_gradients[epoch] is a list where each element stores the gradients for one batch/step. The naming might be a bit confusing, but it does store the per-step gradients as required by the paper.

dattri/algorithm/dvemb.py
```diff
@@ -0,0 +1,237 @@
...
+
+        sorted_epochs = sorted(self.cached_gradients.keys(), reverse=True)
+
+        for epoch in tqdm(sorted_epochs, desc="Computing DVEmb per Epoch"):
+            epoch_embeddings = torch.zeros(
+                (num_total_samples, grad_dim),
+                device=self.device,
+                dtype=grad_dtype,
+            )
+            num_iterations_in_epoch = len(self.cached_gradients[epoch])
```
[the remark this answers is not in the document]
BilyHurington: self.cached_gradients[epoch] is a list of tensors, where each tensor is a batch of gradients. So its length is the number of batches, which equals the number of iterations in that epoch, not the total number of samples.

jiaqima:
Please address the failed tests.

TheaperDeng:
> Please address the failed tests.

We are still working on a performance test to reproduce the a small experiment's result presented on the paper.

dattri/algorithm/dvemb.py
```diff
@@ -0,0 +1,588 @@
...
+
+    DVEmb captures temporal dependence in training by computing data value embeddings
+    that approximate trajectory-specific leave-one-out influence. This implementation
+    stores embeddings for each epoch separately, allowing for epoch-specific analysis.
+    """
+
+    def __init__(
+        self,
+        model: nn.Module,
+        loss_func: Callable,
```
TheaperDeng: We may want to use AttributionTask here for model, loss_func's integration.

dattri/algorithm/dvemb.py
```diff
@@ -0,0 +1,588 @@
...
+
+        if not self.use_factorization:
+
+            @flatten_func(self.model)
+            def vmap_loss_fn(
+                params: Tensor,
+                data_tensors: Tuple[Tensor, Tensor],
+            ) -> Tensor:
+                inputs = data_tensors[0].unsqueeze(0)
+                targets = data_tensors[1].unsqueeze(0)
```
TheaperDeng: If we use AttributionTask (https://github.com/TRAIS-Lab/dattri/blob/main/dattri/task.py#L40) This will be a user defined function and we don't really need to assume user's data is a tuple of 2 tensors.

dattri/algorithm/dvemb.py
```diff
@@ -0,0 +1,588 @@
...
+        Args:
+            model: The PyTorch model to be attributed.
+            loss_func: A per-sample loss function. It should take model parameters
+                       and a single data sample (unbatched) as input.
+            device: The device to run computations on (e.g., "cpu", "cuda").
+            proj_dim: The dimension for projection (if used).
+            factorization_type: Type of gradient factorization to use. Options are
+                                "none"(default),
+                                "kronecker"(same with paper),
+                                or "elementwise"(more memory-efficient).
```
TheaperDeng: I guess this is not directly related to memory since our behavior repect the `proj_dim` parameter as the main parameter to control the memory

dattri/algorithm/dvemb.py
```diff
@@ -0,0 +1,588 @@
...
+                    output_dim = layer.weight.shape[0]
+                    input_dim = layer.weight.shape[1]
+                    project_output = self._generate_projector(
+                        output_dim,
+                        self.projection_dim,
+                    )
+                    project_input = self._generate_projector(
+                        input_dim,
+                        self.projection_dim,
+                    )
```
TheaperDeng: Let's use the projector once https://github.com/TRAIS-Lab/dattri/pull/203 is merged.

dattri/algorithm/dvemb.py
```diff
@@ -0,0 +1,588 @@
...
+                self.projector = self._generate_projector(
+                    num_features,
+                    self.projection_dim,
+                ).to(
+                    device=self.device,
+                    dtype=per_sample_grads.dtype,
+                )
+            projected_grads = per_sample_grads @ self.projector
+
+            scaling_factor = 1.0 / math.sqrt(self.projection_dim)
```
TheaperDeng: We already have a normalization in line 134 right?

dattri/algorithm/dvemb.py
BilyHurington: fixed

dattri/algorithm/dvemb.py
```diff
@@ -0,0 +1,588 @@
...
+    stores embeddings for each epoch separately, allowing for epoch-specific analysis.
+    """
+
+    def __init__(
+        self,
+        model: nn.Module,
+        loss_func: Callable,
+        device: str = "cpu",
+        proj_dim: Optional[int] = None,
+        factorization_type: str = "none",
```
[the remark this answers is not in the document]
BilyHurington: added

dattri/algorithm/dvemb.py
```diff
@@ -31,19 +32,18 @@ class DVEmbAttributor:
 
     def __init__(
         self,
-        model: nn.Module,
-        loss_func: Callable,
-        device: str = "cpu",
+        task: AttributionTask,
+        criterion: nn.Module,
```
gebbing12: Since you have already integrated the `criterion` into the loss_func of `AttributionTask`, it may be better to use get_loss_func to get it instead of redefining it.

dattri/algorithm/dvemb.py
```diff
@@ -0,0 +1,614 @@
...
+    DVEmb captures temporal dependence in training by computing data value embeddings
+    that approximate trajectory-specific leave-one-out influence. This implementation
+    stores embeddings for each epoch separately, allowing for epoch-specific analysis.
+    """
+
+    def __init__(  # noqa: PLR0912, PLR0915
+        self,
+        task: AttributionTask,
+        criterion: nn.Module,
+        proj_dim: Optional[int] = None,
```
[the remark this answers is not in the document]
BilyHurington: https://github.com/TRAIS-Lab/dattri/blob/f5fb64b17c72e1bb7f563367f0d4e0dcb5a85e05/dattri/algorithm/logra.py#L93-L98

I can set a default feature_batch_size (e.g., 64) so that projector_kwargs becomes optional, similar to LoGra. However, I think we should still keep projector_kwargs as an argument because dvemb can often use a larger batch size during attribution, and some users might want to tune this for performance.

dattri/algorithm/dvemb.py
```diff
@@ -0,0 +1,655 @@
...
+                            loss = nn.CrossEntropyLoss()
+                            yhat = model(image.to(device))
+                            return loss(yhat, label.to(device))
+                        ```.
+            proj_dim: The dimension for projection (if used). Defaults to None,
+                      meaning no projection.
+            factorization_type: Type of gradient factorization to use. Options are
+                                "none" (default),
+                                "kronecker" (same as in the paper),
+                                or "elementwise" (better performance while
```
jiaqima: Maybe elaborate a little bit about what "elementwise" does?

dattri/algorithm/dvemb.py
```diff
@@ -0,0 +1,655 @@
...
+        test_grads_tensor = torch.cat(test_grads, dim=0).to(self.device)
+
+        # Select active embeddings (according to epoch if specified)
+        if epoch is not None:
+            if epoch not in self.embeddings:
+                msg = f"Embeddings for epoch {epoch} not found."
+                raise ValueError(msg)
+            active_embeddings = self.embeddings[epoch]
+        else:
+            active_embeddings = torch.stack(list(self.embeddings.values())).sum(dim=0)
```
jiaqima: should we do sum or mean?

dattri/algorithm/dvemb.py
```diff
@@ -0,0 +1,655 @@
...
+                _output: Tensor,
+            ) -> None:
+                a = inputs[0].detach()
+
+                if a.dim() > 2:  # noqa: PLR2004
+                    a = a.reshape(a.size(0), -1)
+                caches[idx]["A"] = a
+                caches[idx]["has_bias"] = layer.bias is not None
+
+            return _hook
```
Jiachen-T-Wang: In Mixed Precision, this hook forces the retention of fp32 input activations. Since Autograd saves a separate bf16 copy for the backward pass, this hook prevents the fp32 tensor from being freed, significantly increasing memory usage. To fix this, we need to use saved_tensors_hooks instead, which is non-trivial (see my recent revision to GhostSuite). For now, I suggest to make a note here and in the README.

VanderLFZ: Hi Jiachen, we checked the issue and found the following. In non–mixed precision training, the caches and autograd share the same storage since they both use fp32. In mixed precision training, during each batch (step), only the cache['A'] at layer[0] does not share storage between the cache and autograd, it stores fp32 while the autograd stores bf16. After that, the behavior is the same as in non–mixed precision training, where in both cases, we use the caches to compute projected_grads and then move them to CPU. Since this process happens for every batch and we do not accumulate them across multiple steps, the peak memory reduction for bf16 mixed precision training would be at most about 50% × N, where N denotes the batch size.

PR #203, opened 2025-10-19
Replace fast_jl with sjlt

sleepymalc:
## Description

Add three projectors (Mask, SJLT, and GraSS) and remove one (FJLT).

> [!WARNING]
> ~Now, we actually pre-compute/sample and store all the randomness when initializing the projectors. This might affect some attributors if the current code initializes a new projector for every projection (i.e., it might lead to OOM if memory is not cleaned up immediately). However, the time complexity will be the same theoretically, with some potential speedup with a rewrite (such that we initialize projectors and reuse them when initializing the attributor).~

Update: Due to the need for ensembling for `ChunkedCudaProjector`, the current implementation falls back to the sampling/compute randomness on-the-fly style. This can potentially be improved with more detailed condition control (e.g., when `CudaProjector` is not initialized via `ChunkedCudaProjector`, then we store pre-computed randomness).


jiaqima:
> This might affect some attributors if the current code initializes a new projector for every projection

I think the API of most (if not all) of our previous projectors involves two steps: 1) initialize the projector (all the randomness happens here): `projection = SomeProjector()`; and then 2) apply (the same) projection to each vector: `x_projected = projection(x)`. 

@TheaperDeng @tingwl0122 might be able to confirm?

sleepymalc:
> > This might affect some attributors if the current code initializes a new projector for every projection
> 
> I think the API of most (if not all) of our previous projectors involves two steps: 1) initialize the projector (all the randomness happens here): `projection = SomeProjector()`; and then 2) apply (the same) projection to each vector: `x_projected = projection(x)`.
> 

For instance, in the previous FJLT in `CudaProjector` and projection-matrix-based methods in `BasicProjector`, due to `ensemble_id` argument used in `project()` (required by chunking if the problem is too large), we need to compute the randomness on the fly.

For instance, if we look at `__init__`, nothing is pre-computed.

https://github.com/TRAIS-Lab/dattri/blob/e5eb9e79e8730c93152f80f93bdc91d3c9f26146/dattri/func/projection.py#L265-L322

And if we look at 
https://github.com/TRAIS-Lab/dattri/blob/e5eb9e79e8730c93152f80f93bdc91d3c9f26146/dattri/func/projection.py#L369

`fjlt` actually takes in an `ensemble_id` and computes a unique seed inside `project`, and initializes randomness on the fly.

I initially want to do pre-computation and then realize that due to `ensemble_id`, it's hard to do so. I have switched to computing randomness on-the-fly design for now.

dattri/func/projection.py
```diff
@@ -40,6 +40,9 @@ class ProjectionType(str, Enum):
 
     normal: str = "normal"
     rademacher: str = "rademacher"
+    sjlt: str = "sjlt"
+    mask: str = "mask"
```
jiaqima: I would suggest first having a PR which adds sjlt and replaces fast_jl ONLY (without having mask and grass).

Note that the current github action tests will skip any tests dependent on GPU, thus all tests related to fast_jl are skipped. So we cannot solely rely on the github action tests for this particular PR, and need to do a few more manual tests to make sure the replacement is successful.

jiaqima: Either works.

jiaqima:
You may want to change the fast_jl related parts in README.md and pyproject.toml as well in this PR.

dattri/func/projection.py
TheaperDeng: Just realize this could happen. Only like half of the test cases are run in our github action and the others are not tested. How about I raise a PR this weekend to fix it first (but it can only be tested on local side)?

```bash
test/dattri/func/test_proj.py::TestBasicProjector::test_basic_projector_shape PASSED [ 72%]
test/dattri/func/test_proj.py::TestCudaProjector::test_project_output_shape SKIPPED [ 74%]
test/dattri/func/test_proj.py::TestChunkedCudaProjector::test_project_output_shape SKIPPED [ 75%]
test/dattri/func/test_proj.py::TestArnoldiProjector::test_arnoldi_projector PASSED [ 77%]
test/dattri/func/test_proj.py::TestGetProjection::test_arnoldi_project PASSED [ 78%]
test/dattri/func/test_proj.py::TestGetProjection::test_basicprojector PASSED [ 80%]
test/dattri/func/test_proj.py::TestGetProjection::test_chunkedcudaprojector SKIPPED [ 81%]
test/dattri/func/test_proj.py::TestGetProjection::test_cudaprojector SKIPPED [ 82%]
test/dattri/func/test_proj.py::TestGetProjection::test_tensor_input_chunked_cuda SKIPPED [ 84%]
test/dattri/func/test_proj.py::TestGetProjection::test_tensor_input_cpu PASSED [ 85%]
test/dattri/func/test_proj.py::TestGetProjection::test_tensor_input_cuda SKIPPED [ 87%]
```

TheaperDeng: Yes I know, maybe we can also add a checklist in the PR template asking our contributors to have a snapshot or something showing the skipped test are passed on their local environment.

jiaqima: For now I think it's ok for us to check the need for local tests manually in our code review.

sleepymalc:
I have updated the `test_proj.py` and locally confirmed that it passes all tests on different GPU (which is something fast_jl struggled with even when correctly installed, if I remember correctly).

test/dattri/func/test_proj.py
```diff
@@ -442,16 +437,15 @@ def test_cudaprojector(self):
...
-        torch.cuda.is_available() and FAST_JL_AVAILABLE,
-        "CUDA is not available or fast_jl is not installed",
+        torch.cuda.is_available() and SJLT_AVAILABLE,
+        "CUDA is not available or sjlt is not installed",
     )
     def test_chunkedcudaprojector(self):
-        """Test funcionality of ChunkedCudaProjector."""
-        test_batch_size = 64
+        """Test functionality of ChunkedCudaProjector."""
+        test_batch_size = 16
```
jiaqima: Is it possible to make this test pass by adjusting the chunk size instead of the scale of the model?

In addition, is there an estimate on the memory overhead of this projection? For example, something like "On A40, we can run forward and backward for a model with num_layers L and batch_size B; but we can only afford a model with num_layers L' and batch_size B'."

sleepymalc: Yes, it's a tricky bug actually. Previously the chunkprojector as well as the parameter_count is not correctly written.

sleepymalc: Yes, it's a tricky bug actually. Previously the `ChunkedCudaProjector` as well as the `parameter_count` is not correctly written.

test/dattri/func/test_proj.py
```diff
@@ -14,13 +14,13 @@
     random_project,
 )
 
-# Check if fast_jl is available
-# TODO: Remove this check once we remove fast_jl dependency
+# Check if sjlt is available
```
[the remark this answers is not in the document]
jiaqima: Same for all the following.

dattri/func/utils.py
```diff
@@ -134,7 +134,7 @@ def _get_parameter_chunk_sizes(
 def get_parameter_chunk_sizes(
     param_shape_list: List,
     batch_size: int,
-) -> tuple[int, int]:
+) -> tuple[int, List[int]]:
```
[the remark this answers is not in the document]
jiaqima: The implementation of these two functions are different though. Might wait for @tingwl0122 's explanation before we delete it.

test/dattri/func/test_proj.py
```diff
@@ -42,29 +42,28 @@ def test_basic_projector_shape(self):
...
-            device="cpu",
+            device=self.device,
         )
 
         test_grads = torch.randn(10, self.feature_dim)
         projected_grads = self.projector.project(test_grads, ensemble_id=0)
         assert projected_grads.shape == (10, self.proj_dim)
 
 
-# TODO: Remove FAST_JL_AVAILABLE check once we remove fast_jl dependency
```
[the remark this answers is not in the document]
sleepymalc: Not sure what's the reason we want to remove this check?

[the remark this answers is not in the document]
jiaqima: In either case, I think we should include sjlt to the test env in pyproject.toml?

dattri/func/projection.py
```diff
@@ -279,47 +274,123 @@ def __init__(
                 (dimension of the gradient vectors).
             proj_dim (int): Dimension we project *to* during the projection step
             seed (int): Random seed.
-            proj_type (ProjectionType): Type of randomness to use for
-                projection matrix (rademacher or normal).
-            device (str): CUDA device to use.
+            proj_type (ProjectionType): The random projection type used for the
+                projection. Available options are "sjlt", "rademacher", "normal".
+            device (torch.device): Device to use. Defaults to cpu.
```
TheaperDeng: In Cudaprojecter the device is not default to cpu. I think we don't really have a default value for device?

dattri/func/projection.py
```diff
@@ -333,50 +404,27 @@ def project(
...
-        except RuntimeError as e:
-            if "CUDA error: too many resources requested for launch" in str(e):
-                # provide a more helpful error message
-                msg = "The batch size of the CudaProjector is too large for your GPU. \
-                    Reduce it by using the proj_max_batch_size argument.\
-                    \nOriginal error:"
-                raise RuntimeError(msg) from e
-            raise e from None
+        elif self.proj_type in [ProjectionType.rademacher, ProjectionType.normal]:
+            result = features @ self.proj_matrix / (self.proj_dim**0.5)
```
TheaperDeng: Correct me if I'm wrong. If we use cudaprojector with non-sjlt types (like rademacher and normal). Th e implementation falls back to a naive projection (w/o block size in basic projector)

dattri/func/projection.py
```diff
@@ -811,77 +864,68 @@ def make_random_projector(
...
         use_half_precision (bool): If True, torch.float16 will be used for all
             computations and arrays will be stored in torch.float16.
 
     Returns:
         The initialized projector object
         (CudaProjector, ChunkedCudaProjector, or BasicProjector).
     """
-    using_cuda_projector = False
-    dtype = torch.float16 if use_half_precision else torch.float32
+    dtype = torch.bfloat16 if use_half_precision else torch.float32
```
TheaperDeng: I am not sure about this dtype. BF16 is a fairly new data type with same range (but less precision) with FP32 and supported in business devices like 3rd edition Xeon, Zen4 and A-series NV GPU and afterwards. (And it is supported by less personal computer devices).

Maybe we can have a logic here:
1) If "cpu" is used, check its support for bf16, if not, then we raise an InputError
2) If "cuda" is used, check its support for bf16, if not we fall back to fp16.

@tingwl0122 correct me if I am wrong. I think the mean idea here for supporting half precision is memory saving rather than computation cost right?

sleepymalc: The current sjlt doesn't support fp16, unfortunately. That's why I changed it to bf16.

sleepymalc: It's mainly because I can't find a way to make cuda kernel to support fp16.

dattri/func/projection.py
sleepymalc: Yes

dattri/func/utils.py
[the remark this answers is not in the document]
sleepymalc: If I remember correctly `_get_parameter_chunk_sizes ` is different from `get_parameter_chunk_sizes ` before this pr. @tingwl0122 correct me if I'm wrong.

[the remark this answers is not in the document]
jiaqima: Yes, let's remove it then.

pyproject.toml
```diff
@@ -29,7 +29,7 @@ dependencies = [
 homepage = "https://github.com/TRAIS-Lab/dattri"
 
 [project.optional-dependencies]
-fast_jl = ["fast_jl"]
+sjlt = ["sjlt"]
 test = ["build", "pytest", "pre-commit", "ruff", "darglint", "scikit-learn", "pretty_midi", "requests", "matplotlib", "pyparsing", "torch>=2.2.0", "torchvision>=0.17.0"]
```
jiaqima: add dependency to sjlt for the test option? In the current version, the tests for sjlt will never be triggered due to the checks of sjlt installation.

dattri/func/projection.py
TheaperDeng: @sleepymalc Do we want to implement this if else logic?

PR #208, opened 2025-11-04
[CI/CD] Add manually triggered CUDA/GPU involved tests

TheaperDeng:
### 1. Motivation and Context

We don't need run CUDA tests for every PR, so we design a manually trigger machanism (through comment "run gpu test").

### 2. Summary of the change

1. Add a new workflow file to trigger such machanism

A trial snapshot is here:
pic

2. Change the default unit test only run non-GPU tasks (by add a mark ("gpu") to GPU required tests)
```bash
...
collecting ... collected 82 items / 10 deselected / 72 selected
...
```

### DIFF
--- a/test/dattri/func/test_proj.py
+++ b/test/dattri/func/test_proj.py
@@ -859,10 +859,6 @@ class TestProjectionCombinationsCPU(unittest.TestCase):
                     self._test_projection_combination("cpu", proj_type, dtype)
 
 
-@unittest.skipUnless(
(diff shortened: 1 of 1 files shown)


TheaperDeng:
Comment: 
- The PR is locally tested in my forked repo. There are some thing I can not validate before merging this PR. @jiaqima 
- Please check if I miss any GPU-required unittest for the gpu mark. @sleepymalc

PR #211, opened 2025-11-05
Fix bug to use self-hosted server

TheaperDeng:
## Description

A quick fix to the gpu triggered unittest.


TheaperDeng:
run gpu test

PR #218, opened 2025-11-13
Improved LoGra with General two-stage compressor

sleepymalc:
## Description

1. Improving gradient hooks to fully avoid the gradient of parameter computation
2. Implement a general two-stage compressor (supporting LoGra and FactGraSS)

### 1. Motivation and Context

1. The current gradient hook still computes the gradient of the parameter, resulting in redundant computation.
2. The compressor is specific to LoGra, and can not integrate nicely with new compressor such as FactGraSS.

### 2. Summary of the change

1. Implement a monkey patch for a custom backward pass that avoids computing full parameter gradients.
2. Implement a general compressor class that handles two-stage compression.

- [x] N/A: No test will be added (please justify)

### DIFF
--- a/test/dattri/algorithm/test_factgrass.py
+++ b/test/dattri/algorithm/test_factgrass.py
@@ -64,17 +64,12 @@ class TestFactGraSSAttributor:
     def test_attribute(self) -> None:
         """Ensure attribution works with two-stage projection."""
         self.attributor.cache(self.train_loader)
-        assert self.attributor.sparsifiers, "Sparsifiers should be initialized"
-        assert self.attributor.projectors, "Projectors should be initialized"
-
-        # Check dimensions: sparsifiers should have intermediate_dim/2 = 256/2 = 16
-        # (per-component dimension), projectors should have proj_dim = 64
+        assert self.attributor.compressors, "Compressors should be initialized"
         assert self.attributor.layer_dims == [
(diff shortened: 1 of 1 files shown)


dattri/algorithm/block_projected_if/__init__.py
```diff
@@ -0,0 +1,24 @@
+"""LoGra utilities package.
```
jiaqima: LoGra -> BlockProjectedIF?

jiaqima (approved):
LGTM. Thanks!

dattri/algorithm/block_projected_if/__init__.py
sleepymalc: Thanks for the catch.

test/dattri/algorithm/test_logra.py
```diff
@@ -4,7 +4,7 @@
 from torch import nn
 from torch.utils.data import DataLoader, TensorDataset
 
-from dattri.algorithm.logra.logra import LoGraAttributor
+from dattri.algorithm import LoGraAttributor
```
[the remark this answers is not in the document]
sleepymalc: Sounds good

sleepymalc:
I think it's ready, but I'll wait for @TheaperDeng to confirm.

PR #220, opened 2025-11-19
Revise README for installation and clarity improvements

sleepymalc:
## Description

Updated installation instructions and improved clarity in the README. Update the environment with CUDA 12.4 and torch 2.6.0.

### 1. Motivation and Context

Update the environment settings with newer CUDA and torch versions. This enables modern packages such as [flash attention](https://github.com/Dao-AILab/flash-attention) (which requires CUDA 12.0).


sleepymalc:
We are ready to merge then?

PR #228, opened 2025-11-29
Fixing batch size for ChunkedCudaProjector

sleepymalc:
## Description

Currently, the `ChunkedCudaProjector.project()` still accepts dict input. However, as we already supported `ChunkedCudaProjector.dict_project`, we should remove this stale support. Additionally, this PR fixes the following:
1. The current logic for `batch_size` detection for tensor input is incorrect, due to the boundary (last) batch size might be different.
2. There is a test constructing ChunkedPCudarojector directly and passing in dictionary input.


TheaperDeng (approved):
LGTM

PR #231, opened 2025-12-31
random projection in influence functions

### LINKED ISSUE #227: Influence Function with Projection
We should support the random projection for the influence function.

haochend413:
## Description

Add random projection for some IF attributors
- IF_Explicit
- IF_EKFAC
- IF_DataInf

Closes #227

### DIFF
--- a/dattri/algorithm/influence_function.py
+++ b/dattri/algorithm/influence_function.py
@@ -1141,6 +1141,13 @@ class IFAttributorEKFAC(BaseInnerProductAttributor):
         ckpt_idx: int,  # noqa:ARG002
         train_rep: torch.Tensor,
     ) -> torch.Tensor:
+        """Transform train representation with projection.
+        Args:
+            ckpt_idx (int): Index of model checkpoints for ensembling.
+            data (Tuple[torch.Tensor, ...]): Training data batch.
+        Returns:
+            torch.Tensor: Projected training representation.
+        """
         # project 
         if self.projector_kwargs and self.input_projectors:
             layer_rep_proj = self._project_rep(train_rep)
--- a/dattri/algorithm/influence_function.py
+++ b/dattri/algorithm/influence_function.py
@@ -5,7 +5,7 @@ from __future__ import annotations
 from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
-    from typing import List, Optional, Tuple, Union
+    from typing import Any, Dict, List, Optional, Tuple, Union
 
     from torch import Tensor
     from torch.utils.data import DataLoader
@@ -18,9 +18,9 @@ from functools import partial
 import torch
 
 from dattri.func.hessian import ihvp_arnoldi, ihvp_cg, ihvp_explicit, ihvp_lissa
-
-from .base import BaseAttributor, BaseInnerProductAttributor
 from dattri.func.projection import random_project
+
+from .base import BaseInnerProductAttributor
 
 
 def _lissa_collate_fn(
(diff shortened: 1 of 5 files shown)


dattri/algorithm/influence_function.py
```diff
@@ -67,17 +69,48 @@ def __init__(
...
+        """Transform the train representations via random projection.
+
+        Args:
+            ckpt_idx (int): Index of the model checkpoints.
+            train_rep (torch.Tensor): Train representations to be transformed.
+
+        Returns:
+            torch.Tensor: Transformed train representations with projected dimension.
+        """
+        from dattri.func.projection import random_project
```
sleepymalc: This seems redundant:

https://github.com/TRAIS-Lab/dattri/pull/231/changes#diff-e756f81a0441b8319db94cc2a72041edee5cee3936e2150a424ada0afb6803bbR21

haochend413:
> Yes, I'm using the LiSSA Attributor, with the CIFAR10 datasets and mobilev2 model. Here is the link to a running file. You can use these checkpoints and datasets attached. Let me know if you need anything else. Thank you! notebook: https://colab.research.google.com/drive/1ccdlWwcp6WhKe2SRIdK8MIRR7MGs_95m?usp=sharing
> 
> checkpoints and datasets: [datasets](https://drive.google.com/file/d/1r2CJ8_VbgUhhoh9CwDnZnRSN_IfFFXPT/view?usp=sharing) [checkpoints](https://drive.google.com/file/d/1tqg31eyyMgfbIAPEO8Use_uUSbgpa2z9/view?usp=sharing)

Thank you for the information! Though I don't think projection is available for LiSSA since hvp for LiSSA is calculated using the torch vjp directly?

dattri/algorithm/influence_function.py
```diff
@@ -642,6 +688,27 @@ def cache(
...
+                layer_sizes[layer_index] += model_params[idx].numel()
+
+            # create projectors for all the layers
+            for layer_idx in range(param_layer_map[-1] + 1):
+                sample_features = torch.zeros(1, layer_sizes[layer_idx])
+                self.layer_projectors[layer_idx] = random_project(
+                    sample_features,
+                    feature_batch_size=1,
+                    **self.projector_kwargs,
+                )
```
TheaperDeng: There is no need to change this code block, but I think the `random_project` may need some change to avoid such kind of cumbersome projector creation. TODO: make this an issue

dattri/func/hessian.py
```diff
@@ -244,11 +245,15 @@ def ihvp_explicit(
...
             `regularization * I`, where `I` is the identity matrix directly added
             to the Hessian matrix.
+        projector_kwargs (Optional[Dict[str, Any]]): Keyword arguments for
+            random projection. Default: None.
 
     Returns:
         A function that takes a tuple of Tensor `x` and a vector `v` and returns
         the IHVP of the Hessian of `func` and `v`.
     """
+    from dattri.func.projection import random_project
```
TheaperDeng: Is it possible to put this import to the header of the file?

TheaperDeng (commented):
Good job. I think we also need unit test to make sure the change works for all three attributors.

dattri/func/hessian.py
```diff
@@ -263,12 +268,27 @@ def _ihvp_explicit_func(x: Tuple[torch.Tensor, ...], v: Tensor) -> Tensor:
             The IHVP value.
         """
         hessian_tensor = hessian_func(*x)
+        if projector_kwargs is not None:
+            sample_features = torch.zeros(1, hessian_tensor.shape[0])
+            projector = random_project(
+                sample_features,
+                1,
+                **projector_kwargs,
```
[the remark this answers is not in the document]
TonyZhou05: Seems like this is also a bug here that `random_project()` call is missing a required parameter proj_max_batch_size
image

PR #244, opened 2026-02-08
Add CLAUDE.md with project conventions

jiaqima:
## Description

Add `CLAUDE.md` to the project root with project conventions and development guide for AI-assisted development.

### 1. Motivation and Context

Given that many of us will be using AI coding assistants when contributing to this project, it might be better to maintain an official `CLAUDE.md` to improve consistency.

### 2. Summary of the change

Added a single `CLAUDE.md` file covering:
- Project overview (purpose, domain)
- Build & development setup (pip install, conda environment)
- Common commands (test, lint, format)
- Code style conventions (Ruff, Google-style docstrings, naming, imports, Python 3.8+ compatibility)
- Architecture overview (algorithm/, func/, benchmark/, metric/, model_util/, task.py)
- Testing conventions (pytest, GPU markers, synthetic data)

- [x] N/A: No test will be added — this is a documentation-only change (no source code modified)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -72,18 +72,20 @@ ruff check --fix
 - GPU tests marked with `@pytest.mark.gpu`
 - Tests use synthetic data (TensorDataset) for speed
 
-## Contributing — Project Structure
+## Contributing 
+
+### Project Structure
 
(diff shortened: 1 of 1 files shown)


sleepymalc:
One thing that is good to add is that we should document more on some styles we want to enforce. E.g., always separate the attributor task and the attributor, always keep the docstring, etc. We can add this later, though.

PR #245, opened 2026-02-10
Tracin group attribute

Soulknight-T:
## Description

Add Tracin test_dataloader_group to support memory efficient computation.
Add example/test of test_dataloader_group.

### 1. Motivation and Context

Add an alternative way to get score matrix with less memory used

### 2. Summary of the change

Add TestDataloaderGroup class to TracIn.py
Add test to test_tracin.py
Add usage example
Update GitHub example workflow

- Unit test: Typically, this should be included if you implemented a new function/fixed a bug.

### DIFF
--- a/dattri/algorithm/tracin.py
+++ b/dattri/algorithm/tracin.py
@@ -178,10 +178,12 @@ class TracInAttributor(BaseAttributor):
                     grad_t = self.grad_target_func(parameters, test_batch_data)
                     if self.proj_params.proj_dim is not None:
                         # define the projector for this batch of data
+                        # use grad_t.shape[0] (not test_batch_data[0]) to support
+                        # DataloaderGroup where test_batch_data is a DataLoader
                         self.test_random_project = random_project(
                             grad_t,
                             proj_params=RandomProjectionParams(
-                                feature_batch_size=test_batch_data[0].shape[0],
+                                feature_batch_size=grad_t.shape[0],
                                 device=self.device,
                                 **self.proj_params.model_dump(),
                             ),
--- a/test/dattri/algorithm/test_tracin.py
+++ b/test/dattri/algorithm/test_tracin.py
@@ -676,7 +676,6 @@ class TestTracInAttributor:
             task=task,
             weight_list=torch.ones(len(checkpoint_list)),
             normalized_grad=False,
-            projector_kwargs=None,
             device="cpu",
         )
 


.github/workflows/examples_test.yml
```diff
@@ -52,11 +52,12 @@ jobs:
...
+        python examples/data_cleaning/tracin_dataloader_group.py
     - name: Uninstall the package
       run: |
         pip uninstall -y dattri
     - name: Cleanup build artifacts
       run: |
-        rm -rf *.egg-info 
-    - uses: eviden-actions/clean-self-hosted-runner@v1 
+        rm -rf *.egg-info
+    - uses: eviden-actions/clean-self-hosted-runner@v1
```
TheaperDeng: I guess this is an automatic change by our pre-commit system? If yes, then we can keep them.

dattri/algorithm/tracin.py
```diff
@@ -98,7 +135,49 @@ def attribute(  # noqa: PLR0912
...
+        ) -> Tensor:
+            test_grads_accumulator = 0
+
+            for sub_batch in test_item:
+                if isinstance(sub_batch, (tuple, list)):
+                    temp = tuple(x.to(self.device) for x in sub_batch)
+                else:
+                    temp = sub_batch.to(self.device)
+
+                sub_grad = torch.nan_to_num(self.grad_target_func(parameters, temp))
```
TheaperDeng: We may want to have a slighly different user-defined loss/target function (only shown in example) and avoid the changes in `attributor.attribute()`

TheaperDeng (changes requested):

examples/data_cleaning/tracin_dataloader_group.py
```diff
@@ -0,0 +1,104 @@
...
+
+    def __init__(self, original_test_dataloader: DataLoader) -> None:
+        """Initialize the DataloaderGroup.
+
+        Args:
+            original_test_dataloader (DataLoader):
+                The PyTorch dataloader for individual test data samples
+        """
+        super().__init__(torch.utils.data.TensorDataset(torch.zeros(1)))
+        self.original_test_dataloader = original_test_dataloader
```
TheaperDeng: self.shuffle = original_test_dataloader.shuffle

dattri/algorithm/tracin.py
```diff
@@ -98,7 +98,10 @@ def attribute(  # noqa: PLR0912
             Tensor: The influence of the training set on the test set, with
                 the shape of (num_train_samples, num_test_samples).
         """
-        _check_shuffle(test_dataloader)
+        if hasattr(test_dataloader, "original_test_dataloader"):
+            _check_shuffle(test_dataloader.original_test_dataloader)
+        else:
+            _check_shuffle(test_dataloader)
```
TheaperDeng: Remove this change.

examples/data_cleaning/tracin_dataloader_group.py
```diff
@@ -0,0 +1,104 @@
...
+        """
+        return 1
+
+
+if __name__ == "__main__":
+    torch.manual_seed(42)
+    input_dim, n_train, n_test = 2, 10, 5
+
+    model = nn.Linear(input_dim, 1, bias=False)
+    model.weight.data.fill_(1.0)
```
TheaperDeng: Use mnist_mlp

examples/data_cleaning/tracin_dataloader_group.py
Soulknight-T: Actually it's not needed

dattri/task.py
```diff
@@ -256,7 +263,34 @@ def get_grad_target_func(
...
+                batches = list(data)
+
+                def flat_group_target(flat_params):
+                    params_dict = _unflatten_params(flat_params, model_ref)
+                    return group_target_func(params_dict, batches)
+
+                g = grad(flat_group_target)(parameters)
+                return g.unsqueeze(0)
+
+            return base_grad_target(parameters, data)
```
TheaperDeng: Please still call `return self.grad_target_func` if `group_target_func=False` and only return a separately wrapped function if `group_target_func=True`

TheaperDeng (changes requested):

Issue #242, opened 2026-02-07
Refactor projector_kwargs across different influence_functions


TonyZhou05:
Issue:
In influence function attributors we do not have projector_kwargs, but instead flatten it into the function signature, E.g proj_dim, seed etc and assemble the kwargs internally. Others such as trak, we are using projector_kwargs as a signature. This introduces inconsistencies in API and creates confusions for users as they explore different interfaces. 

Proposals: 
Since projector_kwargs is part of the subsystem of influence function, we should unify the API by creating a strongly typed config for it (Pydantic class) For different influence functions, we can define/extend different Pydantic types to ensure compatibility.

<img width="283" height="153" alt="Image" src="https://github.com/user-attachments/assets/286851dd-7885-44f2-bfe1-a988a55dc986" />

PR #248, opened 2026-02-17
refactor projector parameters API to enable strongly typed interface …

TonyZhou05:
### 1. Motivation and Context

The projector_kwargs field in attributor is confusing and does not convey explicitly to the user what they can pass in for the projection for customization. This PR is trying to refactor this to improve the user workflow as well as internal reference to different projection parameters.
implemented this task: https://github.com/TRAIS-Lab/dattri/issues/242

### 2. Summary of the change

Deprecate the usage of projector_kwargs and replace with strongly typed ProjectionParam Pydantic classes tailored for different attributors and random projections.

- [ ] ...

### DIFF
--- a/.github/workflows/doc.yml
+++ b/.github/workflows/doc.yml
@@ -15,7 +15,7 @@ jobs:
       - uses: actions/setup-python@v5
         with:
           python-version: ${{ matrix.python-version }}
-          cache: 'pip'  # enable pip cache
+          # cache: 'pip'  # disable pip cache for self-host runner
 
       - name: Install dependencies (binary wheels, no legacy easy_install)
         env:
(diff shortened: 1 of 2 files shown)


dattri/params/projection.py
```diff
@@ -8,12 +8,11 @@
 from pydantic import BaseModel, ConfigDict, field_validator
 
 
-class GeneralProjectionParams(BaseModel):
-    """General projection params used by LoGra, FactGraSS, TracIn."""
+class BaseProjectionParams(BaseModel):
```
TonyZhou05: This is for attributors to inherit from, so they have the flexibility to define whatever proj_dim type they want: proj_dim/proj_dim_per_layer etc. GeneralProjectionParam can inherit from this for more generic purpose such as random_project and parent attributor

sleepymalc: This makes sense.

